PR #14, opened 2020-07-17
Use .nvmrc for picking the right version of Node.js

### LINKED ISSUE #7: Run nvm install as part of install command.
We have `.nvmrc` these days, so picking the right version of npm should be the part of install command (rather than extra lines in install routine in travis file).

kabalin:
This closes #7


stronk7:
Note that, right now, with #13 merged, this is conflicting, apart from any other work still needed here.

stronk7:
Was thinking id, maybe, in order to support older branches up to some extent... we could asume the previous `lts/carbon` (or `v8.16.1`). That, for sure will make happy some older versions.

Just a random thought..

kabalin:
> Was thinking if, maybe, in order to support older branches up to some extent... we could assume the previous lts/carbon (or v8.16.1). That, for sure will make happy some older versions.
Hmm, that it an interesting idea. So, you recon we can substitute v8.16.1 for all installations that do not have .nvmrc in place?

stronk7:
> Hmm, that it an interesting idea. So, you recon we can substitute v8.16.1 for all installations that do not have .nvmrc in place?

Haven’t checked but if I look to the local_codechecker .travis.yml (cannot provide link now, sorry, I’m afk) I see that there I’m using 8/8.9 for 32/33/34 stables, so it may do the trick (defaulting to them), yep.

stronk7:
Link: https://github.com/moodlehq/moodle-local_codechecker/blob/master/.travis.yml#L52

(Also 36, BTW)

kabalin:
Thanks @stronk7, thinking actually to add .nvmrc to cloned repo if it does not exist (defaulting to `v8.16.1`) and proceed with install as usual.

stronk7:
> Thanks @stronk7, thinking actually to add .nvmrc to cloned repo if it does not exist (defaulting to `v8.16.1`) and proceed with install as usual.

Thanks sounds perfect. +1

kabalin:
I added commit to use `lts/carbon` as default if `.nvmrc` is not found. Not sure if we need to add an old branch test to repo. It is test covered and I also tested it here: https://travis-ci.org/github/kabalin/moodle-plugin-ci/builds/710005378 (ignore failure, it is Behat test related), according to output it picks `8.17.0` as expected (https://travis-ci.org/github/kabalin/moodle-plugin-ci/jobs/710005386#L332)

stronk7:
> Thinking if we need CLI option for install step (and/or dedicated ENV var) to skip npm install? Not sure if there is a use case for that though. May be can address it as separate ticket if needed.

I cannot imagine it right now, if current and future versions are going to have {{.nvmrc}} cannot imagine any reason to disable the automatic handling. Let's see, surely we'll discover it once it's released, heh.

BTW, little detail, there is a remaining `-vvv` left in the own tests. About 34_STABLE and behat I don't remember, I've commented @ #20 , though I'm not 100% sure (it sounds to me that auto-rerun is an option somewhere...).

kabalin:
> there is a remaining -vvv left in the own tests

True 😃 Actually, thinking if we should keep it in `.travis.yml` - makes easier to spot issue, verify versions in use, etc.?

stronk7:
> True 😃 Actually, thinking if we should keep it in `.travis.yml` - makes easier to spot issue, verify versions in use, etc.?

That’s ok for me, ok.

docs/CHANGELOG.md
```diff
@@ -22,7 +22,11 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
   this is no longer required, you can remove this step.
 - Updated version of `moodlehq/moodle-local_codechecker` to v2.9.6
 - Updated [.travis.dist.yml] to build Moodle 3.9
-- Updated [.travis.dist.yml] to use nvm 14.0.0
+- `moodle-plugin-ci install` installs Node.js (npm) using the version
+  specified in .nvmrc file. See
+  [#7](https://github.com/moodlehq/moodle-plugin-ci/issues/7)
+- ACTION REQUIRED: You may safely remove `nvm install <version>` and `nvm use
+  <version>` from .travis.yml for Moodle 3.5 and above, this is now a part of installation routine.
```
stronk7, line 29: I'm going to amend this @ #19 (coz now all branches are covered, older too).

kabalin: That't true, thanks!

Issue #20, opened 2020-07-20
Behat re-run fails on server not availalbe.


kabalin:
Here is an example: https://travis-ci.org/github/kabalin/moodle-plugin-ci/jobs/710005386#L2424

Not sure how re-run works, is it triggered as part of same command, or we somehow stop servers by the time it tries again. Need to investigate. It is easy to replicate with build of `moodle-plugin-ci` itself for `MOODLE_34_STABLE`.

kabalin:
Discovered on testing #14

stronk7:
Wow, curious. Only 34_STABLE? And you did nothing to force the automatic rerun? Maybe we used to rerun automatically then? In any case, I think it's all the same script (`run.php`?) the one providing that. From memory behat does not provide that by default (but I may be wrong!).

stronk7:
Ah, I see we are passing the `auto-run` option defaulting to 2. Ok, that explains why it's called. But I cannot imagine why it's failing on rerun, as far as it's the same script (`run.php`) the one handling that. So it should be the same (env-wise) for the original run and any rerun.

Uhm... sure there is an explanation...

kabalin:
> Wow, curious. Only 34_STABLE?
I did not test with any other version. I think it is just `local_travis` fixture incompatibility (no idea why, may be it will become clear on investigation), and it fails on chrome run only.

> And you did nothing to force the automatic rerun?
It defaults to 2 as you found already.

May be I'll add some debugging output and see what happens.

kabalin:
The issue is purely related to Moodle 3.4 used in test, that fails on behat test (using chrome profile) in the the latest `moodle-plugin-ci`. It probably not worth investigating why since 3.4 is not supported and those users who still wish to test can use default profile (firefox) that does not have this issue. Also, we have #15 which will allow passing explicit selenium images (assuming the issue is caused by selenium-chrome driver version in use).

Behat test forcing failure has been run using 3.9 stable to check re-run behaviour (in both Chrome and Firefox):
https://travis-ci.org/github/kabalin/moodle-plugin-ci/jobs/710787644
Re-run was is triggered correctly, there are no "The base URL is not a behat test site" errors shown.

Closing the issue.

PR #22, opened 2020-07-21
Add --node-version parameter to install step.

kabalin:
The option allows to override default Node.js version (which is picked
from .nvmrc) by providing --node-version parameter or defining
`NODE_VERSION` env variable. The value of this parameter should be
compatible with `nvm install` command, e.g. `v8.9`, `8.9.0`,
`lts/erbium`. Basically, assume it is added to .nvmrc and nvm should be
able to consume it while running `nvm install && nvm use`.

### DIFF
--- a/src/Installer/VendorInstaller.php
+++ b/src/Installer/VendorInstaller.php
@@ -38,39 +38,36 @@ class VendorInstaller extends AbstractInstaller
     private $execute;
 
     /**
+     * @var string
+     */
+    public $nodeVer;
+
+    /**
      * Define legacy Node version to use when .nvmrc is absent (Moodle < 3.5).
      *
      * @var string
      */
     private $legacyNodeVersion = 'lts/carbon';
 
-    public function __construct(Moodle $moodle, MoodlePlugin $plugin, Execute $execute)
+    /**
+     * @param Moodle       $moodle
+     * @param MoodlePlugin $plugin
+     * @param Execute      $execute
+     * @param string       $nodeVer
+     */
+    public function __construct(Moodle $moodle, MoodlePlugin $plugin, Execute $execute, $nodeVer)
     {
         $this->moodle  = $moodle;
         $this->plugin  = $plugin;
         $this->execute = $execute;
+        $this->nodeVer = $nodeVer;
     }
 
     public function install()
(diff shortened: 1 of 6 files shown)


kabalin:
Test run using `MOODLE_23_STABLE` branch: https://travis-ci.org/github/kabalin/moodle-plugin-ci/builds/710325106
All works as expected (fails on Behat, but that is a different story #20 ).

kabalin:
Made some refactoring to avoid warnings (it seems like step count method it called before anything else is started).

tests/Installer/VendorInstallerTest.php
```diff
@@ -25,40 +25,45 @@ public function testInstall()
...
+            new DummyMoodle($this->moodleDir),
+            new MoodlePlugin($this->pluginDir),
+            new DummyExecute(),
+            $userVersion
+        );
+        $installer->installNode();
+
+        // Expect .nvmrc containing user specified version.
+        $this->assertTrue(is_file($this->moodleDir.'/.nvmrc'));
+        $this->assertSame(file_get_contents($this->moodleDir.'/.nvmrc'), $userVersion);
```
stronk7, line 67: tiny detail, 1st param in assertions should be the expectation, not the value being checked.

src/Installer/VendorInstaller.php
```diff
@@ -83,7 +80,7 @@ public function install()
 
         $this->execute->mustRunAll($processes);
 
-        $this->getOutput()->step('Install npm dependencies');
+        $this->getOutput()->step('Install nmp dependencies');
```
stronk7, line 83: nmp?

kabalin: Heh, I was trying to modify that string to mention Node.js, but decided to keep as is at the end.

tests/Installer/VendorInstallerTest.php
kabalin: True, fixed.

stronk7:
Have updated #19 to 1) group all the "ACTION REQUIRED" points together and also to add a comment about the --node-version => NODE_VERSION => .nvmrc => lts/carbon fallback system.

PR #24, opened 2020-07-22
Complete the release docs with some important details

stronk7:
- what to change.
- how to apply those changes.
- when to tag.
- what to expect after pushing tags.
- how to proceed if the automatic "deploy" fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ jobs:
     - php: 7.0
       env: MOODLE_BRANCH=MOODLE_35_STABLE
     - stage: Deploy
+      if: tag IS present
       before_install: skip
       install: skip
       script: skip
@@ -83,7 +84,7 @@ jobs:
         provider: releases
         api_key: $GITHUB_TOKEN
         file: build/moodle-plugin-ci.phar
-        cleanup: false
+        skip_cleanup: true
         on:
           repo: moodlehq/moodle-plugin-ci
           tags: true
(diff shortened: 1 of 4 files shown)


kabalin:
@stronk7, I pushed some changes to avoid creating new PR (to experiment with this "allow maintainers to add commits" feature), but then realised you probably can't review them 😃 

Anyway, I have investigated how releases feature work and updated release documentation accordingly. Also I added a condition to deploy stage, so it will not be used unless the build has been triggered by tag.

stronk7:
Yeah, I've been viewing that "Add more commits..." in a number of repos and always have wondered if that really means that the commits are pushed really. And they are! (I've fetched them now and have them locally). I was imagining that maybe some "interim" branch was created or so.

But I think it's nice thing, after all. Specially if only maintainers can do that, LOL. Thanks for testing the feature!

Going to take a look briefly...

.travis.yml
```diff
@@ -73,6 +73,7 @@ jobs:
     - php: 7.0
       env: MOODLE_BRANCH=MOODLE_35_STABLE
     - stage: Deploy
+      if: tag IS present
```
stronk7, line 76: Aren't this and the `tags: true` below redundant?  Note I've not looked docs.

stronk7: Aha, so the condition is to skip the whole stage. So we save ourselves from running `before_deploy` (make) completely.

kabalin: Yes, this `if` condition is to skip whole stage (like you see in the build for this PR). The stage will only be used when build has been triggered by tag (either tag was pushed or created via interface). That `on: tags` is condition for deployment itself, probably it is redundant as stage will never be reached in normal PR or master builds, but I would keep it :)

stronk7: Yeah, makes sense, thanks!

PR #26, opened 2020-08-18
Fix the commit reference of the latest local_moodlecheck plugin

mudrd8mz:
The travis-ci is still using the old version of the local_moodlecheck plugin. As a result, I am seeing some false failures such as "Invalid inline phpdocs tag @see found" which I am not getting when using the latest version locally.

I suspect that the commit c006b90ac2283fed1975f7039ab9cc40c75bdf0c only raised the "version". But that does not really do anything imho without setting the "reference" hash correctly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,7 @@
         "source": {
           "url": "https://github.com/moodlehq/moodle-local_moodlecheck.git",
           "type": "git",
-          "reference": "ceab6b90ff9f4680468bd6720b2add290b608820"
+          "reference": "79fe5956f9e9be906499c01dbc59695c2412643d"
         }
       }
     }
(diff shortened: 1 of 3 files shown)


kabalin:
Hi @mudrd8mz, thanks a lot for contribution. You need to bump version at line 67, otherwise composer will not recognise changes, then run `composer update moodlehq/moodle-local_moodlecheck` to re-generate `composer.lock` and add it to commit. Also mentioning the change in changelog would be appreciated.

@stronk7 shall we make point release following this (opportunity to test release instruction also)?

kabalin:
May be we need to bump moodlecheck following landing of https://github.com/moodlehq/moodle-local_moodlecheck/pull/56, would be useful to have it.

mudrd8mz:
Patch amended.

danmarsden:
ping! - any chance of getting this merged in? - I've updated some of my code to get Green tests in plugins db, but it's getting big red failures in travis-ci

mudrd8mz:
Patch amended again to use the latest version of the `local_moodlecheck` plugin and the change mentioned in the changelog.

stronk7:
I've added a new commit on top of @mudrd8mz one, providing:

- bump to very-latest 79fe5956f9e9be906499c01dbc59695c2412643d (merged mins ago).
- .lock file regenerated with php70 and checking the the commit there matches the above (it was not matching in original commit)

Feel free to merge this squashed, np here.

kabalin (approved):
This looks good to me, thanks @mudrd8mz and @stronk7. I would probably not run composer update for all dependencies for the purpose of moodle-local_codecheck upgrade, but since we are about to release this does not matter.

PR #28, opened 2020-08-31
Grunt should also check .js.map files

marinaglancy:


kabalin (approved):
Thanks @marinaglancy, useful fix 👍 Can you add a statement to changelog as well, so we it will be mentioned as one of the changes for the release?

kabalin (approved):
Looks great to me, thanks @marinaglancy !

PR #31, opened 2020-09-04
Prepare for 3.0.1 release

kabalin:


stronk7 (approved):
Looks ok, have checked all the commits since 3.0 and it seems that everything worth being commented is in the CHANGELOG, so all right.

Will merge once travis finishes... (if nobody does before)

Issue #33, opened 2020-09-04
Change the "unreleased" link to show changes from last (3.0.1) release


stronk7:
Detected immediately after merging #32, we also need to update the "unreleased" link on every release, so it just shows changes from the version being released (3.0.1). Right now it's showing changes since 3.0.0 instead (because we missed to update it).

Not critical, but have to fix it, surely documenting it too, so we don't miss it next time.

stronk7:
Closing this as fixed by #34

PR #34, opened 2020-09-11
Prepare for 3.0.2 release

stronk7:
Just bumping to current local_ci and local_moodlecheck that
come with some nice improvements:

- Support .mustachelintignore in mustache linting.
- Better valid API detection (online and external).
- Anonymous classes false positives.

Only the 1st one is worth being shared in the CHANGELOG, because it's something that devs can now go adding to their plugins, instead of using IGNOREFILES to completely ignore those files.

Note that this PR, if accepted must be applied WITHOUT commit merge ("rebase and merge", usually), because will be tagging it as 3.0.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,11 +52,11 @@
       "type": "package",
       "package": {
         "name": "moodlehq/moodle-local_ci",
-        "version": "1.0.5",
+        "version": "1.0.6",
         "source": {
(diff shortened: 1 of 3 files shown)


kabalin (approved):
All looks good to me, I see you fixed #33 also. Thanks!

stronk7:
Ah, lol, did not remember #33 existed!

PR #35, opened 2020-09-11
Let's always use manual tagging and push upstream

stronk7:
While it's also possible to use GH release interface
it's better to avoid using it. Just tag manually, send
the tag upstream and travis will both:

1) Create the needed release artifacts (moodle-plugin-ci.phar)
2) Perform the official release.


kabalin (approved):
Great, thanks @stronk7

Issue #21, opened 2020-07-20
Use Symfony 4.4 (current LTS)


kabalin:
As Symfony 3.4 support comes to end in 11/2020, we need to switch Symfony packages to current LTS version: https://symfony.com/releases/4.4

stronk7:
Pity about this is the Sympathy 4.4 requires php71. So:

- people will need to move their tests (matrix, jobs) up if they are still using php70.
- minimum compatible version of Moodle will be 3.2 (older don't support php71).

stronk7:
Results with just versions and dependencies updated: https://travis-ci.org/github/stronk7/moodle-plugin-ci/builds/726331973

stronk7:
Closing this is favor of #151 , that is using Symfony 5, already in master branch and near 4.0.0 release.

PR #38, opened 2020-09-15
Bump symfony dep injection component to ^3.4

stronk7:
That makes a number of components to be also updated to 3.4.x
and requires php >7.0.8 (instead of 7.0.0).

Also updated psalm-baseline.xml to ignore 'node-version' potential
wrong assignment values.

Note I've not run `--update-baseline` because it comes with other
changes that I don't want to modify now (removing some options).
I leave that for the experts because I'm not sure how the baseline
is decided and which part is automatic or manual.

A step before #21

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,18 +74,18 @@
     }
   ],
   "require": {
-    "php": ">=7.0.0",
+    "php": ">=7.0.8",
     "moodlehq/moodle-local_codechecker": "^2.9.7",
     "moodlehq/moodle-local_ci": "^1.0.5",
     "moodlehq/moodle-local_moodlecheck": "^1.0.2",
     "sebastian/phpcpd": "^3.0",
     "phpmd/phpmd": "^2.2",
-    "symfony/dotenv": "^3.3",
-    "symfony/filesystem": "^3.3",
-    "symfony/finder": "^3.3",
-    "symfony/console": "^3.3",
-    "symfony/yaml": "^3.3",
-    "symfony/process": "^3.3",
+    "symfony/dotenv": "^3.4",
+    "symfony/filesystem": "^3.4",
+    "symfony/finder": "^3.4",
+    "symfony/console": "^3.4",
+    "symfony/yaml": "^3.4",
+    "symfony/process": "^3.4",
     "php-parallel-lint/php-parallel-lint": "^1.2.0",
     "php-parallel-lint/php-console-highlighter": "^0.5",
     "psr/log": "^1.0",
@@ -99,7 +99,7 @@
   },
   "config": {
     "platform": {
-      "php": "7.0"
+      "php": "7.0.8"
     }
   },
   "autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "7c9944fe5c7563822cc23d5310a40bf8",
+    "content-hash": "ad1d86c385702f0cd0caae16d4213850",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -74,12 +74,13 @@
...
   ],
   "require": {
-    "php": ">=7.0.0",
+    "php": ">=7.0.8",
     "moodlehq/moodle-local_codechecker": "^2.9.7",
     "moodlehq/moodle-local_ci": "^1.0.5",
     "moodlehq/moodle-local_moodlecheck": "^1.0.2",
     "sebastian/phpcpd": "^3.0",
     "phpmd/phpmd": "^2.2",
+    "symfony/dependency-injection": "^3.4",
```
kabalin, line 83: So, it is not actually bump, but adding a new dependency?

kabalin (approved):
> I'm not sure how the baseline
is decided and which part is automatic or manual.
Not sure either, let's leave that for experts 😄 Otherwise looks good.

composer.json
stronk7: Yeah, without that, it defaults to the 3.3.6 current version... because it’s installed as sub.dependency and some rule out there leads to that.

So I went the explicit way, adding it as 1.level dependency, so the rest only can adapt.

Maybe the same can be achieved if, instead of adding that, we bump everything else to ^3.4, because effectively that’s what is happening (one 3.4 forces all the others to be also 3.4).

I can try that way, if you thin it makes more sense, I just wanted to leave the json as unmodified as possible, but maybe has more sense to explicitly bump all to 3.4 and then the dependency one (I bet) will come also with its 3.4 matching version.

stronk7: Confirmed, I've deleted the `dependeny-injection` line and, instead, bumped everything else to ^3.4 and it leads to the creation of the very same composer.lock (zero differences):

```$ git diff
diff --git a/composer.json b/composer.json
index 05f7471..832c3b8 100644
--- a/composer.json
+++ b/composer.json
@@ -80,13 +80,12 @@
     "moodlehq/moodle-local_moodlecheck": "^1.0.2",
     "sebastian/phpcpd": "^3.0",
     "phpmd/phpmd": "^2.2",
-    "symfony/dependency-injection": "^3.4",
-    "symfony/dotenv": "^3.3",
-    "symfony/filesystem": "^3.3",
-    "symfony/finder": "^3.3",
-    "symfony/console": "^3.3",
-    "symfony/yaml": "^3.3",
-    "symfony/process": "^3.3",
+    "symfony/dotenv": "^3.4",
+    "symfony/filesystem": "^3.4",
+    "symfony/finder": "^3.4",
+    "symfony/console": "^3.4",
+    "symfony/yaml": "^3.4",
+    "symfony/process": "^3.4",
     "php-parallel-lint/php-parallel-lint": "^1.2.0",
     "php-parallel-lint/php-console-highlighter": "^0.5",
     "psr/log": "^1.0",
diff --git a/composer.lock b/composer.lock
index df7e292..b92ccf5 100644
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "677e9ec4531b99f98923230fa5c09b6c",
+    "content-hash": "ad1d86c385702f0cd0caae16d4213850",
     "packages": [
         {
             "name": "composer/ca-bundle",
```

So, maybe it's better to jump all them explicitly together? And the `dependeny-injection` will come automatically?

kabalin: Heh, you changed already. Well, either way is OK for me to be hones, but explicit bumping makes the change more intuitive for other developers. Let't merge the final version, thanks @stronk7

stronk7: Done, have replaced the commit by the "all up to ^3.4 one.

kabalin (approved):
+1

composer.json
kabalin: Great, thanks!

stronk7:
Thanks!

PR #41, opened 2020-10-05
Add JS snippet to fixture mustache template to test JS linting.

kabalin:

PR #42, opened 2020-10-14
Prepare for 3.0.3 release

vmdef:
Hello @sarjona,

This is an update to use the latest version of local_codechecker.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,11 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 ## [Unreleased]
 No unreleased changes.
 
+## [3.0.3] - 2020-10-16
+### Changed
+- Updated version of [moodle-local_codechecker](https://github.com/moodlehq/moodle-local_codechecker) to v2.9.8.
+- Updated project dependencies ([moodle-local_ci](https://github.com/moodlehq/moodle-local_ci)).
+
 ## [3.0.2] - 2020-09-11
 ### Added
 - Skip HTML validation in mustache templates adding a `.mustachelintignore` standard ignores file to the plugin. Useful for templates containing specific syntax not being valid HTML (Ionic..).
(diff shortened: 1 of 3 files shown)


docs/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-No unreleased changes.
+- Updated composer.json to use version 2.9.8 of the `local_codechecker` plugin.
```
stronk7, line 12: For some reason I think I prefer to say:

    - Updated version of `moodlehq/moodle-local_codechecker` to v2.9.7

Note this is exclusively for `local_codechecker`, because we produce real versioned releases. For others like `local_ci` or `local_moodlecheck`, that don't have real releases, surely it's ok to say, simply:

    - Updated project dependencies (`local_ci`, `other_important_component`...)

I'm not strong on that, just sharing my opinion, because the "Updated composer.json to use..." doesn't sound great in my brain.

Opinions?

stronk7 (commented):
Apart from the comment about try to normalize how we list in the changes the updated dependencies... this looks ok.

Let's decide about that tiny point and this can progress, I think, thanks!

docs/CHANGELOG.md
vmdef: It's ok for me.  So the rule would be to write "Updated version of..." when there are real versions, "Updated project dependencies..." otherwise, right?

stronk7: For me it makes sense, yep. Let's see if anybody chimes in with other idea and let's go that way on Monday, if that's ok for you.

stronk7:
Ah, apart from the codechecker... but somehow related to releasing a new version of moodle-plugin-ci, just noting that last week we also fixed an [annoying problem @ moodle-local_ci](https://github.com/moodlehq/moodle-local_ci/pull/226) that was causing the mustache-linter to fail when it contained JS code to check.

So, maybe it's not crazy to "abuse" this PR to also bump the moodle-local_ci dependency to current HEAD, so we get that issue (and some others) fixed here.

docs/CHANGELOG.md
kabalin: > Updated project dependencies (`local_ci`, `other_important_component`...)
I agree, we update dependencies.

stronk7: Thanks @kabalin,

so, with 3 of us agreeing, sold:

1) for versioned / key dependencies `Updated version of xxxx to yyy`
2) for all the non-versioned / general dependencies `Updated project dependencies (optional list of key ones)`

vmdef:
Is it ok if I change this PR title for "Prepare for 3.0.3 release" with the local_ci and codechecker updates?

stronk7:
> Is it ok if I change this PR title for "Prepare for 3.0.3 release" with the local_ci and codechecker updates?

Oh, sure!

vmdef:
I had to kick my branch several times, but the changes for 3.0.3 are already there. It is pending #41, shall we include it in this release?

stronk7:
> It is pending #41, shall we include it in this release?

#41 should be now fixed by https://github.com/moodlehq/moodle-local_ci/pull/226 (that's one of the reason we are making a new release here, to incorporate that fix.

Surely we can confirm that after merging this, then rebase @kabalin one (that will force travis to run again, and then look to logs to see if the `"ESLint couldn't find the plugin "eslint-plugin-babel"` warnings when running mustache are gone.

That should be enough, IMO.

stronk7:
Note to reviewer, don't forget to "rebase and merge" this one, instead of normal "create a merge commit" one, because this is going to be release commit.

stronk7 (approved):
It looks perfect, +1 here

PR #46, opened 2020-10-19
Check for a yui/src directory when running the Grunt YUI task

### LINKED ISSUE #45: grunt task fails with "Unable to find local grunt" (moodle-local_chatlogs specific)
For some reason, moodle-plugin-ci (v3) is having problems with "moodle-local_chatlogs", failing with grunt not available error when running the grunt task). See my "test" branch failing:

https://travis-ci.com/github/stronk7/moodle-local_chatlogs/builds/190839966

I've triple checked [moodle-local_chatlogs' .travis.yml](https://github.com/stronk7/moodle-local_chatlogs/blob/e0e9fd1f3f4fac4863144b37e7d23abc15136564/.travis.yml), comparing with other plugins that are passing and also with the docs and I've run out of ideas. Everything looks ok there.

Also, i've executed moodle-ci-install with `-vvv` to get the installation details and see if there is something strange there:

https://travis-ci.com/github/stronk7/moodle-local_chatlogs/builds/190841587

^^^ Apparently, grunt is installed ok, more yet `grunt ignorefiles` is executed ok, but, later, `mooldle-plugin-ci` fails with "Unable to find local grunt" message.

Need a couple of eyes, sure there is something obvious, but I'm failing to find it, grrr!

Ciao :-)

andrewnicols:
This commit solves an issue in some repositories where the following
message is shown:

    Unable to find local grunt

This message is a little confusing because the Grunt uses both a global,
and a local copy of the Grunt executable. The global variant comes from
the `npm install grunt-cli` Installation task, whilst the local variant
is a dependency of Moodle.

The message actually occurs when the global executable is unable to find
the local variant from the current working directory.

To give an example, given the following directory:

    /home/travis/build/moodle/local/chatlogs/yui/src

Grunt will check each of the parent directories until it finds a
`Gruntfile.js` and/or relevant `node_modules/.bin/grunt` executable, i.e.:

    /home/travis/build/moodle/local/chatlogs/yui/src
    /home/travis/build/moodle/local/chatlogs/yui
    /home/travis/build/moodle/local/chatlogs
    /home/travis/build/moodle/local
    /home/travis/build/moodle
    /home/travis/build
    /home/travis
    /home
    /

If none is found then the "Unable to find local grunt" message is then
shown.

In the issue where this fault was detected, the plugin does have a `yui`
directory, but it uses the legacy module structure where the YUI module
is placed directly into the directory, and not within a Shifted module
structure. That is to say that the following is the actual directory
structure of the plugin:

```
└── yui
    ├── jabberaliases
    │   └── jabberaliases.js
    └── keyboard
        └── keyboard.js
```

This is a perfectly valid YUI structure, but Grunt is unaware of it, and
its use has not been recommended for many years. There is no `src`
directory present, and each module is just in it's own directory outside
of the src structure.

The Grunt Command was configured to perform the following when
processing a `yui` task:

```
case 'yui':
    $yuiDir = $this->plugin->directory.'/yui';
    if (!is_dir($yuiDir)) {
        return null;
    }

    return new GruntTaskModel($task, $yuiDir.'/src', 'yui/build');
```

Essentially it assumes that if the `yui` directory exists then the task
should pass a workingDirectory for `yui/src`.

In this case there is no yui/src directory, but the full path is passed
as a working directory to the `ProcessBuilder`.  The ProcessBuilder does
not care that the directory does not exist (presumably because it may
not exist during build, but may do prior to execution), but the result
is that `grunt` is called with a non-existent working directory.

It is not possible to traverse anywhere where the working directory does
not exist, and so the local Grunt cannot be found.

The solution here is to always look for `yui/src` within the plugin
directory.

Grunt is not aware of the older-style directories and does not perform
any checks on those so there is no loss of functionality.

Fixes #45

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,11 +34,11 @@
       "type": "package",
       "package": {
         "name": "moodlehq/moodle-local_codechecker",
-        "version": "2.9.7",
+        "version": "2.9.8",
         "source": {
           "url": "https://github.com/moodlehq/moodle-local_codechecker.git",
           "type": "git",
-          "reference": "v2.9.7"
+          "reference": "v2.9.8"
         },
         "autoload": {
           "classmap": [
@@ -52,11 +52,11 @@
       "type": "package",
       "package": {
         "name": "moodlehq/moodle-local_ci",
-        "version": "1.0.6",
+        "version": "1.0.7",
         "source": {
           "url": "https://github.com/moodlehq/moodle-local_ci.git",
           "type": "git",
-          "reference": "07ac69779414a3838ee9e95aebbd5068938e16f8"
+          "reference": "a0993ac0d4e9c28a234480f9ed503609e0bac0a8"
         }
       }
     },
@@ -75,7 +75,7 @@
   ],
   "require": {
     "php": ">=7.0.8",
-    "moodlehq/moodle-local_codechecker": "^2.9.7",
+    "moodlehq/moodle-local_codechecker": "^2.9.8",
     "moodlehq/moodle-local_ci": "^1.0.5",
     "moodlehq/moodle-local_moodlecheck": "^1.0.2",
     "sebastian/phpcpd": "^3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ad1d86c385702f0cd0caae16d4213850",
+    "content-hash": "a01d6edb2b56aa7a10f7f63354b2ca45",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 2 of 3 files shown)


stronk7 (changes requested):
Hi Andrew,

great investigation, for sure I'd spent days looking to that!

Only suggestion that I can imagine is to add this fix to the "Fixed" section of the "[Unreleased](https://github.com/moodlehq/moodle-plugin-ci/blob/master/docs/CHANGELOG.md#unreleased)" `docs/CHANGELOG.md` file. Something like (100% up to you, heh):

> - Avoid bogus _"Unable to find local grunt"_ errors with plugins using legacy YUI directories. See #46 for details.

(the link/see is optional, we really don't use to link to issues/prs unless they are some breaking/requiring action changes).

andrewnicols:
Rebased on latest release, and added a note to the changelog.

Thanks Eloy :)

stronk7:
Thanks @andrewnicols ! I've taken the liberty of amending your commit following the [Keep a changelog](https://keepachangelog.com/en/1.0.0/) recommendation, adding the missing "Fixed" section and slightly rewording the issue.

Issue #48, opened 2020-11-09
Explore GitHub Actions as alternative to Travis


kabalin:
As GitHub Actions is free for [public open source repos](https://github.com/features/actions#pricing-details), I suggest to consider making it a default CI option for plugin developers who use `moodle-plugin-ci` (also given GitHub is a popular choice among community plugin developers).

For Travis free usage withdraw discussion, please refer to #47.

Possible steps requred (some may require a separate issue to capture work):
- [x] Proof-of-concept how `moodle-plugin-ci` can be run using GitHub Actions (and support all test steps, docker, etc. no free tier limitations of tools that can be used)
- [x] Moving `moodle-plugin-ci` itself to GitHub Actions (test and release steps, updating release manual) (#54, #60)
- [x] Creating `gha.dist.yml` and adding documentation explaining its content and usage (#72)
- [ ] Adding instruction on wiki (similar to https://docs.moodle.org/dev/Travis_integration).

kabalin:
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

timhunt:
OK folks, I have sort-of made it work:

https://github.com/moodleou/moodle-filter_embedquestion/blob/github-actions/.github/workflows/ci.yml

generated this:

https://github.com/moodleou/moodle-filter_embedquestion/runs/1433460106?check_suite_focus=true (which errored when it found a bug in my plugin.)

It required one small change in this project https://github.com/timhunt/moodle-plugin-ci/commit/41ef4b614870fc880f46865d9cfd577b6f7f05ad. That is, don't try to install grunt globally with Node. Just install it locally, and the run it using npx. That seems like generially a good idea.

That is the only reason why the familiar `composer create-project` command line has been replaced by separate git checkout and composer install.

Obviously, this needs a lot of cleaning up, but I think that is worth doing.

timhunt:
By the one, someone pointed me at https://ropensci.org/technotes/2020/11/19/moving-away-travis/ today. Some interesting background there.

mackensen:
@timhunt Marvellous, you're a couple of hours ahead of me. I thought https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works was interesting reading as well.

timhunt:
Note: related forum thread https://moodle.org/mod/forum/discuss.php?d=413252

[...]

PR #51, opened 2020-11-20
Don't install grunt globally. Instead run it with npx

timhunt:
This is necessary (or at least helpful) for running on github actions.

It also seems like a good idea generally, to make thing more self-contained.

I have not tested to see if this is OK on Travis, but I assume it is.

### DIFF
--- a/src/Installer/VendorInstaller.php
+++ b/src/Installer/VendorInstaller.php
@@ -76,7 +76,7 @@ class VendorInstaller extends AbstractInstaller
         if ($this->plugin->hasUnitTests() || $this->plugin->hasBehatFeatures()) {
             $processes[] = new Process('composer install --no-interaction --prefer-dist', $this->moodle->directory, null, null, null);
         }
-        $processes[] = new Process('npm install -g --no-progress grunt', null, null, null, null);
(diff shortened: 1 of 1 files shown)


scara:
Cross-ref: https://github.com/moodlehq/moodle-plugin-ci/issues/48

HTH,
Matteo

kabalin:
It should not be a problem if you use `uses: actions/setup-node@v1` as part of action. I will respond in #48 soon with my example.

stronk7:
Hi @timhunt , would you, please, {{--amend}} and  force-push your branch, so we check if travis-ci.com takes on the PR now?

timhunt:
It seems to be running even though I did not do anything: https://travis-ci.com/github/moodlehq/moodle-plugin-ci/builds/201315968. Looks like it will finish in ~ 5 mins.

stronk7:
> It seems to be running even though I did not do anything...

Ah, no. That one was an unrelated one (old) that I relaunched manually to see if the plan was in effect. I was asking for your PR to be amended and resent to see if the PRs automations also are back.

timhunt:
Ah! RIght. Done now.

stronk7 (approved):
Thanks Tim, looks perfect!

PR #56, opened 2020-11-27
PHPUnitCommand: Disable memory limit for phpdbg run

kabalin:
This is to prevent memory exhausted error when test coverage is
requested (reflecting dc46266419).


stronk7 (approved):
Thanks @kabalin , this looks ok!

PR #64, opened 2021-01-21
Bump to current local_ci and local_moodlecheck

stronk7:
Not much to say, just bump to current ones to verify they pass and to have the latest fixes applied.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ build/php-cs-fixer.phar:
 
 # Update download URL from https://github.com/vimeo/psalm/releases
 build/psalm.phar:
-	curl -LSs https://github.com/vimeo/psalm/releases/download/3.0.17/psalm.phar -o build/psalm.phar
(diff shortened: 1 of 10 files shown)


stronk7:
My +1 to your fixes, @kabalin , thanks! It had been hours for me to understand (in fact I still don't understand it completely, heh).

kabalin:
psalm 4.x reports too many errors (mostly related to function return type missing) and also have php7.3 as minimum requirement, thus using latest 3.x. Potentially we can have it as composer dependency (rather than fetching using curl) in future.

kabalin:
> My +1 to your fixes, @kabalin , thanks! It had been hours for me to understand (in fact I still don't understand it completely, heh).

No problem at all 😉

kabalin (approved):
The version bump part looks OK to me.

stronk7:
(self-merging, so I can continue with codechecker bump now, thanks!)

PR #66, opened 2021-01-24
Use a matrix to check a range of Moodle versions

james-cnz:
I think it may be useful to check against a range of Moodle versions.  I've copied from this config:
https://github.com/catalyst/moodle-auth_saml2/blob/master/.travis.yml
I don't really understand this, but it seems to work for me (aside from the Behat tests, but that seems to be a separate issue).

Regarding the specific choice of tests to run:
I've included Moodle versions 3.4 and up because these are the versions for which the checks seem to work for me, but commented out Moodle versions that are no longer supported, or not released yet.
I've done most checks with MySQL rather than PostgreSQL since this (or the related MariaDB) is the DB included with the downloadable Moodle installer packages.  I've added PostgreSQL checks for the Moodle LTS versions, since I guess if any Moodle versions deserve extra attention, it's probably these ones.
PHP versions are spread out across these checks, so that each PHP version gets checked at least once.
Not sure if there are better options, this way seemed as good as any to me.

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -13,21 +13,47 @@ cache:
     - $HOME/.composer/cache
     - $HOME/.npm
 
-php:
- - 7.2
- - 7.3
- - 7.4
-
-env:
- global:
-  - MOODLE_BRANCH=MOODLE_39_STABLE
- matrix:
-  - DB=pgsql
-  - DB=mysqli
+matrix:
+ include:
+  # Moodle 3.4
+  #- php: 7.0 # 7.0-7.2
+  #  env: MOODLE_BRANCH=MOODLE_34_STABLE NODE_VERSION=8.9 DB=mysqli
+  # Moodle 3.5
+  - php: 7.0 # 7.0-7.2
+    env: MOODLE_BRANCH=MOODLE_35_STABLE NODE_VERSION=14.15 DB=mysqli
+  # Moodle 3.5, PostgreSQL
+  - php: 7.0 # 7.0-7.2
+    env: MOODLE_BRANCH=MOODLE_35_STABLE NODE_VERSION=14.15 DB=pgsql
+  # Moodle 3.6
+  #- php: 7.0 # 7.0-7.3
+  #  env: MOODLE_BRANCH=MOODLE_36_STABLE NODE_VERSION=8.17 DB=mysqli
+  # Moodle 3.7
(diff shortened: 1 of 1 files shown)


andrewnicols:
Personally I do not think that we should include versions of Moodle which are not officially supported as stable releases. At most I would extend this to security/LTS releases. I would certainly not advise we include default support for 3.4, which went out of security support in May 2019.

james-cnz:
Thanks for your swift response.  :-)

Regarding whether to include checks for only stable Moodle versions, or also security-only supported Moodle versions, I think it makes sense to include the latter, because I think, ideally, it would be good if plugin developers could continue to offer security support for their plugins on Moodle versions that still have security support.  I guess though, these could be commented out, as they may not be relevant to the config in the master/main branch.

Regarding whether to include checks for Moodle version that are no longer supported at all, just checking, I have commented out these versions, are you saying it would be best to not include them at all, even commented out?  I figured that since the checks seemed to work, they may as well be included as an option, but I guess perhaps it makes sense to drop support for them in the CI tool, since support for them is dropped generally.

kabalin:
Hello @james-cnz, thanks for the patch. I am not sure we need complex default version of travis file, it should only contain starting point configuration that works for most people. Using matrix for different versions is [documented](https://github.com/moodlehq/moodle-plugin-ci/blob/master/docs/TravisFileExplained.md), someone who needs more complex configuration can refer to it. I would suggest to keep it configured to current stable only for simplicity.

Other than that, node is currently (since [3.0](https://github.com/moodlehq/moodle-plugin-ci/blob/master/docs/CHANGELOG.md#300---2020-07-23)) configured automatically based on `.nvmrc` in Moodle root directory, so you don't need `NODE_VERSION` env variable.

james-cnz:
Hi @kabalin,

I've done a quick survey of the top 30 Moodle plugins (by number of sites) to get a rough idea of what would work best for most people.  Two of these haven't been updated to support Moodle 3.10 or 3.9, so I've considered obsolete, of the remaining 28:
Latest version supports only Moodle 3.10 or 3.9: 10
Latest version supports multiple Moodle versions, but not 3.6 or <=3.4: 5
Latest version supports multiple Moodle versions, including 3.6 or <=3.4: 13

So by this, the majority (18/28) would benefit from support for multiple Moodle versions, but only a minority, although quite a large one (13/28) would benefit from support for NODE_VERSION.

PR #67, opened 2021-01-25
Use legacy Firefox for legacy php-webdriver

andrewnicols:
I'm pretty sure that this is correct but I don't know how best to test it.

This will be required for MDL-66979

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,9 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
   to avoid memory exhaused errors.
 - Updated project dependencies to current [moodle-local_moodlecheck](https://github.com/moodlehq/moodle-local_moodlecheck) and [moodle-local_ci](https://github.com/moodlehq/moodle-local_ci) versions.
 
+### Added
+- Detect existence of legacy php-webdriver, and use a different Firefox image when it is in use
+
 ## [3.0.3] - 2020-10-16
 ### Changed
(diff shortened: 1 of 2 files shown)


stronk7:
> but I don't know how best to test it.

For sure current tests do verify that your patch continues working with legacy/old firefox, hence, it's detecting the "instaclick" pretty well. That's good!

In any case, it's possible to use the `MOODLE_REPO` and `MOODLE_BRANCH` env variable to point to another repo/branch. Maybe you can do that in 66679_verify branch and report the results of your push commit here. That would be enough, IMO. 

Ciao :)

stronk7:
For the records, I've tried with this commit on top of Andrew's patch and pointing to his core repo clone and dev branches:

https://github.com/stronk7/moodle-plugin-ci/commit/eda401d103cfc868cda6ba8d0da97bceffcb295d

And in this travis execution: https://travis-ci.com/github/stronk7/moodle-plugin-ci/builds/214425537 ,  you can see clearly how 39, 310, 311 and master are using the new firefox image (3.141.59) and older branches are using the old/legacy firefox image (2.53.1).



As expected, so once MDL-66979 lands, moodle-plugin-ci will continue working with all branches (chrome, firefox).

Finally, note that for sure there are some improvements to do here once the new W3C drivers are available, like using the drivers directly (without Selenium) or running headless. Both things are possible with those new drivers and all the machinery in core. But that's not for this issue.

Ciao :-)

kabalin (approved):
The change looks good to me, thanks Andrew! Seems work correctly (per @stronk7 test above).

PR #69, opened 2021-01-26
Re-add coverage param to phpunit test

kabalin:
This also makes changes:

* `make test-phpunit` is removed from GHA workflow (already run as part of `make validate`)
* Use ubuntu 16.04 so we have phpdbg 7.0 available (needed for 3.5)
* Add verbosity.
* Makefile changes: running `make init` is required before running any test/validate command.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Build PHAR
         run: |
-          make build/moodle-plugin-ci.phar
+          make build
           php build/moodle-plugin-ci.phar list
 
       - name: Create Release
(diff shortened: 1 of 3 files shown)


stronk7:
Hi,

I bet there are multiple approaches to this. So your PR is far better than existing, indeed (talking about the `Makefile` refactor, the coverage side of the issue is ok!).

Only thing that called my attention is that init/update are super similar and made me think if maybe we can make the later to also require the former (check-init).

And then, as far as `vendor/autoload.php` is only used once (init)… ditch it and put it in the caller (init)

Up to you, again, there isn't an unique, correct, solution. So I'm happy already, but had to comment the above.

kabalin:
Thanks @stronk7, make sense, updated per your suggestion.

stronk7 (approved):
Thanks @kabalin , have been playing with it locally and everything looks better now. I've not been able to test the original issue (coverage), but both travis and gha have (in self-tests and integration tests). So all right!

PR #73, opened 2021-01-27
GHA: nvm is available.

kabalin:
By the look of it, `nvm` is included in ubuntu environments (e.g. https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md). Not sure how it was overlooked 🤷‍♂️

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,11 +66,6 @@ jobs:
     - name: Check out repository code
       uses: actions/checkout@v2
 
-    - name: Install node
-      uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


stronk7:
Hope it is, heh, because we are [using it in core](https://github.com/moodle/moodle/blob/fd840ab59cd990738fcb692a515ee41838655d5e/.github/workflows/push.yml#L16-L18) since some months ago. 😄

stronk7 (approved):

stronk7:
Makes sense, merging, thanks!

PR #80, opened 2021-02-02
Fix nvm availability check for GHA and update docs.

kabalin:


kabalin:
See https://github.com/moodlehq/moodle-plugin-ci/runs/1813883630?check_suite_focus=true#step:6:75 to confirm that node installation takes place now.

stronk7 (approved):
Makes sense, hence... here we go!

PR #81, opened 2021-02-02
Enable all branches, both for push and pull requests

stronk7:
The most important it the push one, because anybody
contributing to mooodle-plugin-ci wont get their
dev/proposal branches tested. Only if they work on
master, they will.

About pull requests, this is less critical, as far as
master is out only current branch. But tomorrow we may
rename it, or create new upstream branches, so there
isn't anything wrong opening the doors to that now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,6 @@
 name: Moodle Plugin CI
 
-on:
-  push:
-    branches: [ 'master' ]
-  pull_request:
-    branches: [ 'master' ]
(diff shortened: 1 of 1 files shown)


stronk7:
So this is being checked as a pull request... and also the push of the `all_pr_and_push` branch itself to my clone:

https://github.com/stronk7/moodle-plugin-ci/actions/runs/531621675

hence, I can test it before making this PR.

Ciao :-)

kabalin (approved):

Issue #58, opened 2020-12-16
add-plugin action should not assume the default branch of a plugin is called 'master'


timhunt:
https://github.com/moodlehq/moodle-plugin-ci/blob/master/src/Command/AddPluginCommand.php

If --branch option is not given on the command-line (e.g. `moodle-plugin-ci add-plugin --branch main moodleou/moodle-filter_embedquestion`) then this plugin passes the branch name 'master' in the git clone command it runs. In this case, it would be better if it could just not add that argument, so that the install would use the default branch, whatever it is called (e.g. main).

kabalin:
Thanks @timhunt, this has been addressed in the latest release.

Issue #70, opened 2021-01-26
Consider adding opt-in support to dockerhub based on env vars / secrets


stronk7:
Pretty much like it was done @ moodle-docker for travis (we were hitting the limits there).

https://github.com/moodlehq/moodle-docker/pull/150

We should do the same here for:

- own moodle-plugin-ci CI (travis and gha).
- recommended (for plugins) templates (travis and gha). Plus document.

Ciao :-)

PS: Note I've remembered about this because[ I've hit the limits with travis](https://travis-ci.com/github/stronk7/moodle-plugin-ci/jobs/475536614) with `moodle-plugin-ci`, and it was just after 10-12 pushes, so it can happen with relative frequency.

stronk7:
I was curious about some intense / repetitive / big matrix runs I've performed lately with GHA, because they were not hitting DockerHub limits. And, the very same runs, via TravisCI, are hitting them.

So, after a bit of searching, I found this:

https://github.com/actions/virtual-environments/issues/1445#issuecomment-713861495

Basically it seems that Github reached and agreement with Docker and no limits apply to public containers being used by GitHub runners. If private containers or self-hosted runners are used, then credentials may be needed.

So that explains why "normal" jobs aren't hit, so this issue is less critical than I initially was thinking.

Still we have to decide if we want to give any support for those 2 groups (private containers and self-hosted runners).

Ciao :-)

PR #83, opened 2021-02-03
AddPluginCommand should not use master branch by default.

kabalin:
As default branch could be anything, don't explicity set master if `--branch` param is omitted (#58).

Also improves tests:
* `DummyProcess` class has been replaced with mocked `Process`
* `DummyExecute` now properly initialised with output, etc, so we can test what command line is passed to process


stronk7 (approved):
Looks ok, so approved. Waiting for self-tests to finish green...

stronk7:
Ok, I'm merging this, no matter travis is failing because of #70.

PR #84, opened 2021-02-06
Run grunt task using npx.

kabalin:
It looks like we relied on system installed one so far in grunt task (see https://github.com/kabalin/moodle-plugin-ci/runs/1845524756?check_suite_focus=true#step:7:53 modified run that included `whereis grunt`):
Screenshot

Something I forgot to check when #51 was landed 😞 Only noticed as I tested the latest `moodle-plugin-ci` within `moodlehq/moodle-php-apache:7.2` based container in Gitlab and it failed with `sh: 1: grunt: not found`.


stronk7 (approved):
Thanks @kabalin , certainly we didn't look properly in #51 , nice you've found it. I've looked for more potential cases of direct `grunt` execution and haven't found any more, so this looks ok.

PR #88, opened 2021-02-17
Bump local_moodlecheck and switch to composer2

stronk7:
This comes with https://github.com/moodlehq/moodle-local_moodlecheck/pull/69 and, also, together, switched to composer 2 to generate the lock file.


kabalin (approved):
Looks great, thanks.

PR #89, opened 2021-02-18
Test against multiple Moodle branches by default

james-cnz:
This replaces #66.

In response to feedback:
- Moodle versions that are no longer supported at all have been removed entirely, and only current stable Moodle versions are enabled by default, others are commented out.
- The NODE_VERSION env variable is removed.

Also, corresponding changes are made to the new GitHub Actions script, and the docs explaining the Travis and GitHub Actions scripts.

I thought there were enough changes to warrant a new pull request.

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -13,17 +13,29 @@ cache:
     - $HOME/.composer/cache
     - $HOME/.npm
 
-php:
- - 7.2
- - 7.3
- - 7.4
-
-env:
- global:
-  - MOODLE_BRANCH=MOODLE_39_STABLE
- matrix:
-  - DB=pgsql
-  - DB=mysqli
+jobs:
+ include:
+  # Moodle 3.5
+  #- php: 7.0 # 7.0-7.2
+  #  env: MOODLE_BRANCH=MOODLE_35_STABLE DB=mysqli
+  # Moodle 3.8
+  #- php: 7.1 # 7.1-7.4
+  #  env: MOODLE_BRANCH=MOODLE_38_STABLE DB=mysqli
+  # Moodle 3.9
+  - php: 7.2 # 7.2-7.4
+    env: MOODLE_BRANCH=MOODLE_39_STABLE DB=mysqli
+  # Moodle 3.9, PostgreSQL
+  - php: 7.3 # 7.2-7.4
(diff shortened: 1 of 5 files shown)


kabalin:
Hi @james-cnz, thanks a lot for your suggestion and PR. I am not quite sure why we need all currently supported versions coverage in usage manual and dist versions of CI maifests. My opinion, we need to keep `*dist.yml` simple and general, having just one current stable version there as example and also in order to minimise changes we need to make after each Moodle release. There is no purpose to make moodle plugin compatible with all supported Moodle version by default, it is left to descretion of developer to decide which version it should be tested against and modify `.yml` file accordingly. For someone adding CI to their plugin for the first time it is fair to assume they likely need a current stable release as minimum, so we have it listed in both dist version of CI manifests.

Anyway, that is my opinion. What @stronk7 and @mudrd8mz think?

mudrd8mz:
Well, this reflects two different approaches to maintaining and versioning a Moodle plugin.

* Some developers tend to copy the same scheme that Moodle core uses, and they have same `MOODLE_xyz_STABLE` branches for their plugins. For these developers, it makes sense to use the current default GHA workflow and use one plugin branch in various combination of DBs and PHP versions.
* Other developers tend to keep the main (master) branch of their plugin as the "current" one, supporting a few recent Moodle versions. This can be e.g. all versions from the latest LTS version or so. For these developers, it makes sense to have the CI environment more granularly defined, e.g. as is suggested in this pull request.

For my own projects, I use the later scheme. Yet I don't think there are any particular reasons why we should switch to it as the default one. What I believe is enough, is that we have the config for it already present in docs/GHAFileExplained.md

And I agree with arguments and reasoning mentioned by Ruslan above. The current default config is easier to maintain.

stronk7:
Hi,

I think I also prefer to keep it simple. Plus, the "matrix" alternative strategies are already commented in the Docs:

- For travis: https://moodlehq.github.io/moodle-plugin-ci/TravisFileExplained.html
- For GHA: https://moodlehq.github.io/moodle-plugin-ci/GHAFileExplained.html

Personally I also use that matrix alternative in my plugins, testing against various combinations of php/db/moodle, but I feel it falls away from the default (template, simple, easy to maintain) example. So my -1 to flipping the current default.

Thanks for your proposal, @james-cnz ! Much appreciated!

james-cnz:
OK, thanks for considering.

Issue #94, opened 2021-05-08
Add php 8.0 support


stronk7:
Right now it seems that php 8.0 is not well supported by moodle-plugin-ci

1) I tried bumping a plugin ([local_codechecker](https://github.com/moodlehq/moodle-local_codechecker/pull/145)) to 8.0 and started getting problems on install.
2) moodle-plugin-ci own tests (make, travis/gha...) are also failing badly.

So this is about to try to get everything working and passing with php 8.0. Sure that a bunch of dependencies will have to be updated too (phpunit, cs-fixer...) some only dev-dependencies but still.

Maybe, also, worth considering if we should drop support for 35_STABLE (and older) as far as it's (LTS) finished upstream (since May 10th, 2021).

This can be considered part of the [php80 epic](https://tracker.moodle.org/browse/MDL-70745).

Once fixed, we should release a new version ASAP. 

Ciao :-)

ewallah:
With GitHub Actions code coverage is working in PHP 8.0 and PHP 7.4 in master and MOODLE_311_STABLE branches when 
- [using](https://github.com/ewallah/moodle-availability_language/blob/main/.github/workflows/main.yml#L54): `coverage: xdebug`
- [removing](https://github.com/ewallah/moodle-availability_language/blob/main/.github/workflows/main.yml#L69): `phpdbg -d memory_limit=-1 -qrr` from ci/src/Command/PHPUnitCommand.php file

In branches MOODLE_310_STABLE on PHP 7.4 and MOODLE_39_STABLE on PHP 7.3 the code coverage is ignored with the `Error:         Incorrect whitelist config, no code coverage will be generated.`

Perhaps this approach can shed some light on the segmentation faults?

ewallah:
The segmentation faults when generating code coverage seems to be in the CodeUnitCollection class:

`SebastianBergmann\CodeUnit\CodeUnitCollection::mergeWith(): Argument #1 ($filename) must be of type string, SebastianBergmann\CodeUnit\CodeUnitCollection given, called in vendor/phpunit/phpunit/src/Util/Test.php on line 654`

andrewnicols:
> The segmentation faults when generating code coverage seems to be in the CodeUnitCollection class:

Don't use `phpdbg`. The segfault is not an issue on our end, but in phpdbg.

[...]

PR #95, opened 2021-05-08
Bump php-cs-fixer to v2.19.0

stronk7:
That version supports PHP8 and is the last one
before jumping to v3 that comes with huge changes.

This will help in the progress towards PHP8 support happening @ #94


stronk7:
Aha, there is a new rule [no_alias_language_construct_call](https://cs.symfony.com/doc/rules/alias/no_alias_language_construct_call.html) in the Symfony rules. I think it's ok to keep it enabled, so fixing the `die()`uses.

kabalin (approved):
All looks good, thanks @stronk7

stronk7:
Thanks!

PR #97, opened 2021-05-09
Raise PHP min requirements to 7.2 and update PHPUnit to 8.5

stronk7:
This pull request updates PHPUnit (own tests) to version 8.5.x because it's the very first supporting PHP 8.

Also, explicitly we raise general requirements and declare PHP 7.0.x and 7.1.x unsupported. Both are very old PHP versions (EOL > 1y ago) and still, keeping support for PHP 7.2, all recent Moodle branches (34_STABLE and up) continue being supported by the tool (no matter that many of them (34, 35, 36, 36, 38) are out of support).

Still, if somebody needs a version of the tool working with PHP 7.0 or 7.1, they can continue using current `3.0.8` without too much trouble, just they won't be able to add PHP 8.0 at the same time.

Recommendation (to decide): Let's call to the next version `3.1.0` or `3.5.0` or `4.0.0, given there are  noticeable changes in the supported versions.

Finally, a lot of components (composer) are updated with the bump of the min. PHP version, and some small adjustments have been applied to keep everything working. Plus all the PHPUnit 8.5 specific changes.

Another step towards PHP8 support happening @ #94

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
   citest:
     name: CI test
     needs: selftest
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
 
     services:
       postgres:
@@ -53,8 +53,6 @@ jobs:
           - php: '7.4'
             moodle-branch: 'MOODLE_38_STABLE'
           - php: '7.2'
-            moodle-branch: 'MOODLE_35_STABLE'
-          - php: '7.0'
             moodle-branch: 'MOODLE_35_STABLE'
 
     steps:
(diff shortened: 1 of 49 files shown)


kabalin:
I totally agree regarding major version bump for release. Can you rebase it please @stronk7, so codefixer commit is not added.

stronk7:
Sure, now that #95 has been merged, going to rebuild this, pointing to the new incoming version... will comment once ready...

stronk7:
Ok, I've been doing some more tests... and with a couple of issues fixed in codechecker/moodlecheck/ci... I think that we can make moodle-plugin-ci current v3 to run php8 tests... still testing... but just setting "max_input_vars=5000" may be enough.

Of course, that doesn't enable own tests to pass with php8, for that we need this issue and surely others... 

But, I think that it's better to:

1) Verify that v3 goes ok with php8 and perform some minor release with all the detected problems solved. So people will be able to switch to php8 without changing of version.

2) Continue working towards v4, maybe in another branch, adding this issue and any other needed towards the bump in requirements and self-testing with php8 available.

How does that sound?

For now... I'm going to keep this on hold... and concentrate on verifying that a minor is enough (aka, point 1). So we can make a minor release providing that.

scara:
> How does that sound?

👍 

HTH,
Matteo

stronk7:
I'm going to close this in favour of #151, where the new moodle-plugin-ci, with raised requirements, will be planned.

PR #99, opened 2021-05-12
Update project dependencies

stronk7:
Most noticeble one being local_moodlecheck v1.0.7 that
comes with PHP8 specific fixes.

Also updated nikic/php-parser to newer version, able
to run with any >= 7.0 PHP version (8.0 included) and
parse every PHP version (5.x to 8.0).

Changes to the php-parser required some changes in code:

- Look for assignments inside extra Expression element.
- Add 3 psalm cases to baseline, about some casts (the
  objects have the _toString() method, but psalm isn't
  aware).

(details @ https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-4.0.md)

Finally, some changes because of phpmd bump were also required.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,11 +58,11 @@
       "type": "package",
       "package": {
         "name": "moodlehq/moodle-local_moodlecheck",
-        "version": "1.0.6",
+        "version": "1.0.7",
         "source": {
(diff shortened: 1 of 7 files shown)


kabalin (approved):
All look good to me, thanks @stronk7 for working on this.

PR #101, opened 2021-05-13
Move local_ci dependency to tagged one

stronk7:
No much to say, this could have happened in the previous PR, when we also switched to tagged `local_moodlecheck`, but there were other impeding changes there.

So doing it here, now local_ci has tags too, so using them.

Ciao :-)


kabalin (approved):
Great!

Issue #105, opened 2021-06-15
Rename plugin build directory variable to a generic approach


adpe:
Hi @stronk7 

It's somehow confusing, when you are creating a CI/CD configuration file for GitLab and you must set the following variable `TRAVIS_BUILD_DIR` so that the "Install plugins" command will not fail.

Would be good to rename stuff, as not only Travis CI exists out there. Will make a PR soon.

Greets
Adrian

stronk7:
Closing this now that #110 has been merged, thanks!

PR #106, opened 2021-06-15
Rename TRAVIS_BUILD_DIR env variable to CI_BUILD_DIR

adpe:
Will solve #105


kabalin (changes requested):
Please make sure that `TRAVIS_BUILD_DIR` is still supported along with `CI_BUILD_DIR`, so we don't break things for people who use it already in numerous places (like myself). Eventually this can be removed when some other breaking changes will be on horizon.

adpe:
Thanks @kabalin

I was already thinking about that ;-) But maybe it's better if you would have a look on that? For sure my code proposal will not fit your expectations.

adpe:
Here we go @kabalin

kabalin (changes requested):
The code changes look good to me, thanks @adpe . Can you please squash commits into the single one.

PR #108, opened 2021-06-16
docs: Add file ignoring configuration example for GHA.

kabalin:
Rendered change: https://github.com/kabalin/moodle-plugin-ci/blob/ignore-doc/docs/IgnoringFiles.md


stronk7:
Nice work, @gjb2048 & @kabalin !

With #109 now merged... can this be close or still has a point?

BTW, note that both pull requests seem to conflict about how the global travis example is defined @  `IgnoringFiles.md` (just in case we want to, also, progress this).

gjb2048:
@stronk7 No problem.  I changed IgnoringFiles.md at the same time as @kabalin without realising it.  Entirely up to you what version you'd like.

kabalin:
I think it worth keeping this one too (just rebased it), it updates `docs/IgnoringFiles.md` to reflect how this should be configured in GHA.

stronk7 (approved):
Looks ok, good job!

PR #117, opened 2021-07-16
Bump tools dependencies to new, just tagged, ones.

stronk7:
Note that since some months ago we are using tags for all
no matter only codechecker has proper releases, we also
added them to moodlecheck and ci.

Someday we may automate this step and open to allow automatic
updates within some constraints (eg. X.Y.*, ^X.Y....). Not now, anyway.


kabalin (approved):
👍

Issue #93, opened 2021-03-25
injectPHPUnitFilter() needs review for Moodle 3.11 and up


stronk7:
Was looking to some recents builds of moodle-plugin-ci itself when I saw this in output:

```
 RUN  PHPUnit tests for local_ci
Moodle 3.11dev (Build: 20210319), c1db56b501f94b4b9d0e1d0f85487370c637c3e3
Php: 7.4.15, pgsql: 9.6.18, OS: Linux 4.15.0-1077-gcp x86_64
PHPUnit 9.5.2 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 48:
  - Element 'filter': This element is not expected.

  Test results may not be as expected.
....                                                                4 / 4 (100%)

Time: 00:00.185, Memory: 62.00 MB
2450
2451OK (4 tests, 12 assertions)
```
(source: [Travis run](https://travis-ci.com/github/moodlehq/moodle-plugin-ci/jobs/492395880))

Cause for this being that, with Moodle 3.11 and up, we have moved to PHPUnit 9.5 that has replaced the "filter" section by a new "coverage" (better) one. Also the old "whitelist" is now "include" and "exclude" in a descendent of "coverage", not of "include" anymore.
(source: [Moodle Docs](https://docs.moodle.org/dev/Writing_PHPUnit_tests#Upgrading_unit_tests_to_work_with_Moodle_3.11_and_up_.28PHPUnit_9.5.29))

So we need to find a way to, dynamically, provide different "filter/coverage" sections depending of the Moodle version/branch used.

This can affect to both the `phpunit.xml.dist`file and `MoodlePluginCI\Installer\TestSuiteInstaller::injectPHPUnitFilter()`

Ciao :-)

PR #120, opened 2021-07-20
issue #93: injectPHPUnitFilter() review

ewallah:
Time to move from <filters> into <coverage>?

### DIFF
--- a/src/Installer/TestSuiteInstaller.php
+++ b/src/Installer/TestSuiteInstaller.php
@@ -151,8 +151,10 @@ class TestSuiteInstaller extends AbstractInstaller
         $subject   = file_get_contents($config);
         $count     = 0;
 
-        // Replace existing filter.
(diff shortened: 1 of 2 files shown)


stronk7:
Hi,

maybe we should perform the change based on the branch we are running PHPUnit? Only that way we can keep 310_STABLE and older (PHPUnit <= 8.5) working with filter/whitelist and 311_STABLE and up (PHPUnit >= 9.5) with the new coverage/include.

I haven't had a chance to try it myself... but... commands seem to have access to the branch via `this->moodle->getBranch()` and, with that information we can easily keep both alternatives working.

Ciao :-)

ewallah:
HI,

Good point, I only wanted to move things a bit forward.

But perhaps we should also consider it is time to stop removing good configurations and only inject when there is no configuration provided?

R

kabalin:
Closing as being addressed in b1e901ac2d which just landed via #141, thanks for your effort guys!

stronk7:
Closing this now that #141 has been merged

PR #127, opened 2021-08-07
Remove ubuntu 16.04 from self-tests, out of support

stronk7:
And being removed from the GHA setup-php

https://github.com/shivammathur/setup-php
https://github.com/shivammathur/setup-php/issues/452

So, jumping to 20.04 ~and also adding some mising php80 runs.~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   selftest:
     name: CI test (make validate)
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - name: Check out repository code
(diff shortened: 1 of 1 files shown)


stronk7:
OT: had to rerun the GHA workflow because one of the jobs failed (crash). It seems that there is some randomness right now.

See https://github.com/stronk7/moodle-plugin-ci/actions?query=branch%3Abye_ubuntu_16_04++ , it happened the same (different branches affected) while preparing the PR (I did some changes disabling coverage to see if it was related or no).

At the end, it passes... but not nice... going to rerun the workflow a number of times.

stronk7:
Yeah, have run it 3 times:

- 1 passed
- 1 failed the php80 master job.
- 1 failed the php80 master and 311 jobs.

I'm going to remove the new php80 jobs from the matrix, so this will be, exclusively, about to move to 20.04

kabalin (approved):
Thanks @stronk7, looks good to me.

PR #132, opened 2021-09-06
Support subplugins in install

micaherne:
The current implementation of the install command doesn't work as expected where the extra-plugins directory contains a plugin and a subplugin of that plugin. There is an implied dependency between the subplugin and the parent plugin which isn't accounted for, so the subplugin may be installed first, giving the following error:

> In Moodle.php line 113:                       
> The component metadatafieldtype_file has an unknown plugin type of metadatafieldtype

This can be reproduced by cloning the [local_metadata](https://github.com/PoetOS/moodle-local_metadata) plugin into the extra-plugins directory, and then moving the directories within the fieldtypes directory (which are subplugins) up to the root of the extra-plugins directory.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,7 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-No unreleased changes.
+### Added
+- Support for subplugins in the extra-plugins directory for install.
 
 ## [3.2.1] - 2021-07-30
 ### Changed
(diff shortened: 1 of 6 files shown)


micaherne:
I'd like to fix whatever is causing the Travis failures for this pull request, but it's not clear to me what they are as I don't see any clear error report in the output. Can anyone help me understand what the problem is?

kabalin:
> I'd like to fix whatever is causing the Travis failures for this pull request, but it's not clear to me what they are as I don't see any clear error report in the output. Can anyone help me understand what the problem is?

Did you try `make test-fix` locally? Not sure why it fails, might be it that it re-geneate CLI.md and finds differences for some reason.

kabalin (approved):
Hi @micaherne, thanks for the feature, I just added a small commit containing changes following `make test-fixer`. All good to be merged.

PR #135, opened 2021-11-02
GitHub no longer supports the git:// protocol

jonof:
Mentioned at https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git
I believe as I write this GitHub is in the midst of the first scheduled brown-out.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -844,7 +844,7 @@ Moodle repository to clone
 * Accept value: yes
 * Is value required: yes
 * Is multiple: no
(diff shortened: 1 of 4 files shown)


andrewnicols (approved):

stronk7:
Good one, thanks @jonof (et all!)

stronk7:
For the records, I've created #137 to ensure we do some release with these changes rolled before the due date (21 January 2022).

PR #139, opened 2021-11-26
Bump to PHPUnit 6.5.x

stronk7:
While this is also a very outdated PHPUnit version, it aligns
better with current PHP 7.0 min requirement for the whole
moodle-plugin-ci.

Also, better go gradually moving up so transitions to recent
versions 7.x, 8.x, 9.x... are simpler, whenever we decide to
raise moodle-plugin-ci PHP required version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,7 @@
     "padraic/phar-updater": "^1.0"
   },
   "require-dev": {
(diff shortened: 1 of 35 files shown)


kabalin (approved):
Totally agree, thanks for PR!

stronk7:
Thanks!

PR #140, opened 2021-12-03
Unpin Selenium version for Firefox

aspark21:
Aligning with version tested in moodlehq/moodle-docker#195


stronk7 (approved):
Thanks @aspark21,

that's perfect. Soon (once we have a couple of PRs elsewhere applied) we will be also able to start considering the new Selenium 4 versions... let's see.

PR #141, opened 2021-12-03
Coverage php has precedence

### LINKED ISSUE #93: injectPHPUnitFilter() needs review for Moodle 3.11 and up

stronk7:
This has to be considered / merged after #139 (it's built on top of it).

    Revamped phpunit code coverage analysis
    
    - Whenever a tests/coverage.php file is filter, don't apply
      code coverage defaults any more. That file is used by phpunit
      init to properly configure the desired coverage. This rule
      only applies when running tests against Moodle 3.7 and up.
    - Whenever we are using PHPUnit 9.5 and up, use the new coverage
      syntax where <filter><whitelist> became <coverage><include>. This
      rule only applies when running tests against Moodle 3.11 and up.
    - Covered with tests.
    
    Closes #93 and #120

Also:

    Remove the 'header_comment' rule from php-cs-fixer
    
    It was enforcing us to keep the Blackboard header for all files
    when new ones shouldn't be subject to that copyright notice.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -10,16 +10,6 @@
  * License http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
-$header = <<<'EOF'
-This file is part of the Moodle Plugin CI package.
-
-For the full copyright and license information, please view the LICENSE
-file that was distributed with this source code.
-
-Copyright (c) 2018 Blackboard Inc. (http://www.blackboard.com)
-License http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
-EOF;
-
 return (new PhpCsFixer\Config())
     ->setRiskyAllowed(true)
     ->setRules([
@@ -33,7 +23,6 @@ return (new PhpCsFixer\Config())
                                                        'expectedExceptionMessage',
                                                        'expectedExceptionMessageRegExp',
                                                    ]],
-        'header_comment'                        => ['header' => $header],
         'heredoc_to_nowdoc'                     => true,
         'no_extra_blank_lines'                  => ['tokens' => [
                                                        'break', 'continue', 'extra', 'return',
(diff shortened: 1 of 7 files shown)


kabalin (approved):
Thanks a lot @stronk7! Also, well spotted that header_comment thing, should not be there definitely.

stronk7:
Thanks!

PR #149, opened 2022-01-12
Bump to codechecker v3.0.5

stronk7:
Amending the tests to the new Sniffs available. Plus adding a note
to the ExecuteTest.php because it generates output to php logs and
always confuses me, looking why it's happening.


kabalin (approved):
👍

PR #158, opened 2022-02-26
Rebuild AMD test expectations using the new minify/rollup

stronk7:
In [MDL-73915](https://tracker.moodle.org/browse/MDL-73915) NodeJS was bumped to 16 LTS and, also, the
minify system was modernized (all branches), explicitly removing support for IE (master only).

That has caused the resulting built AMD files to be different,
so we have to put here the new versions in order to get self
tests passing.

Note that the 4.0 and up minified versions are different from
the previous ones, and, although they are interchangeable, we
only can assert one of them here. So we go with the 4.0 and up,
explicitly disabling the check for older branches (they are sort
of past, no matter they are still supported).

Also, we have already added as valid the MOODLE_400_STABLE branch
in the GHA workflow, so we won't need to add it in some weeks when the
branch is created (on release).


stronk7:
Self-merging, to be back to green.

Issue #37, opened 2020-09-15
Use utf8mb4 for mysql


marinaglancy:
there is inconsistency in the collation used for testing on mysql in core and moodle-plugin-ci:

.travis.yml creates mysql db with collation utf8mb4_bin - https://github.com/moodle/moodle/blob/master/.travis.yml#L159

moodle-plugin-ci creates it with collation utf8_general_ci - https://github.com/moodlehq/moodle-plugin-ci/blob/master/src/Installer/Database/MySQLDatabase.php#L27

PR #159, opened 2022-02-28
Use utf8mb4 for mysql and mariadb

ewallah:
Moodle 4.0 contains language strings that need the full 4 bytes UTF-8 code. So the MySQL and MariaDB moodle-plugin-ci databases should also support full unicode (issue #37).

### DIFF
--- a/gha.dist.yml
+++ b/gha.dist.yml
@@ -20,6 +20,9 @@ jobs:
         env:
           MYSQL_USER: 'root'
           MYSQL_ALLOW_EMPTY_PASSWORD: "true"
+          MYSQL_CHARACTER_SET_SERVER: "utf8mb4"
+          MYSQL_COLLATION_SERVER: "utf8mb4_unicode_ci"
+
         ports:
           - 3306:3306
         options: --health-cmd="mysqladmin ping" --health-interval 10s --health-timeout 5s --health-retries 3
--- a/src/Installer/Database/MySQLDatabase.php
+++ b/src/Installer/Database/MySQLDatabase.php
@@ -24,7 +24,7 @@ class MySQLDatabase extends AbstractDatabase
         $passOpt  = !empty($this->pass) ? ' --password='.escapeshellarg($this->pass) : '';
         $user     = escapeshellarg($this->user);
         $host     = escapeshellarg($this->host);
-        $createDB = escapeshellarg(sprintf('CREATE DATABASE `%s` DEFAULT CHARACTER SET UTF8 COLLATE utf8_general_ci;', $this->name));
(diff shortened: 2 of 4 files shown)


stronk7:
Just guessing if we shouldn't be using, as commented in #37 (that is fixed by this) if we shouldn't be using the binary collation, that is case-sensitive, instead of any of the case-insensitive ones (_ci).

In core we have some tests that require case-sensitiveness (that enables "case-insensitive" functions to be tested). But the opposite (test case-sensitive stuff with a case-insensitive database) is not possible.

It produces consistent ordering (of course, not using any language special sorting, but pure binary one) and normally it's better for some tests, as commented above.

Thoughts? Ciao :-)

jrchamp:
As far as I can tell, utf8mb4_unicode_ci is the collation that core uses/recommends when Moodle is installed, so the tests should probably be using the same schema definition. Also, it's still possible to perform case-sensitive comparisons using `CONVERT()` and the `BINARY` keyword.

ewallah:
I detected the problem when using the tool_customlang_utils::checkout function in Moodle_40. A dml_write_exception is thrown without utf8mb4.

```
dml_write_exception: Error writing to database (Incorrect string value: '\xF0\x9F\x91\x8B' for column `moodle`.`phpu_tool_customlang`.`original` at row 1
INSERT INTO phpu_tool_customlang (lang,componentid,stringid,original,master,timemodified,outdated,local,timecustomized) VALUES(?,?,?,?,?,?,?,?,?)
[array (
  0 => 'en',
  1 => '495000',
  2 => 'welcomeback',
  3 => 'Welcome back, {$a->firstname}! 👋',
  4 => 'Welcome back, {$a->firstname}! 👋',
  5 => 1642067673,
  6 => 0,
  7 => NULL,
  8 => NULL,
)])
```
I guess plugins will also add icons in the language files.

stronk7:
Hi @ewallah,

yeah, yeah... to switch to `utf8mb4` charset is out of discussion. It's a must and it will be done here.

The ony point I'm not 100% convinced is the collation to use. One side there is the `_bin` (case sensitive) ones that it's the recommended for development and for core to get all tests passing. And other side there is/are the `_unicode_ci / _general_ci` (case insensitive) that are more "real" in sorting terms, but less in others.

So this is just a "theoretical" discussion about which collation should we use with moodle-plugin-ci (the "developers" one, or the "production" one).

Ciao :-)

stronk7:
I've been asking within the team and nobody showed any strong opinion.

So I think it's ok to use `utf8mb4_unicode_ci` (that is the recommended one in general) and done. If somebody/some plugin really needs a case-sensitive database, always can run the tests with other databases or whatever.

As said, in core that's needed just because of a few tests that require binary comparisons by default. That's the reason we always run phpunit with CS/AS collations in all DBs.

So, going to proceed with this as is, thanks @ewallah (and @jrchamp !).

Ciao :-)

PR #160, opened 2022-03-18
fix symfony/process

geichelberger:
When using the moodle-plugin-ci with a gitlab-ci pipeline and Kubernetes, the pod gets not destroyed after the pipeline finishes because there is still a process running. The process is still running because of a signaling issue that does not occur if the first argument is an array. https://github.com/symfony/symfony/pull/21474


stronk7 (approved):
Thanks @geichelberger, this looks correct. Merging...

PR #161, opened 2022-03-18
Add --db-port option

geichelberger:
Add support for non-default database ports. For example, you cannot run a MariaDB and MySQL database on the same host with the default ports.

### DIFF
--- a/res/template/config.php.txt
+++ b/res/template/config.php.txt
@@ -11,7 +11,9 @@ $CFG->dbname    = '{{DBNAME}}';
 $CFG->dbuser    = '{{DBUSER}}';
 $CFG->dbpass    = '{{DBPASS}}';
 $CFG->prefix    = 'mdl_';
-$CFG->dboptions = [];
+$CFG->dboptions = [
+    'dbport' => '{{DBPORT}}',
+];
 
 $CFG->wwwroot  = '{{WWWROOT}}';
 $CFG->dataroot = '{{DATAROOT}}';
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Added
+- `moodle-plugin-ci install` now provides an option `--db-port` to define a custom database port.
 
 ## [3.2.6] - 2022-05-10
 ### Changed
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -813,7 +813,7 @@ Install everything required for CI testing
 
 ### Usage
 
-* `install [--moodle MOODLE] [--data DATA] [--repo REPO] [--branch BRANCH] [--plugin PLUGIN] [--db-type DB-TYPE] [--db-user DB-USER] [--db-pass DB-PASS] [--db-name DB-NAME] [--db-host DB-HOST] [--not-paths NOT-PATHS] [--not-names NOT-NAMES] [--extra-plugins EXTRA-PLUGINS] [--no-init] [--node-version NODE-VERSION]`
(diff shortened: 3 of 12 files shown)


kabalin:
Hi @geichelberger, thanks for your contribution. Please reflect changes in docs/CHANGELOG.md also.

stronk7:
HI,

I just found myself coding exactly the same changes performed by @geichelberger here.

The reason is that, with Moodle 4.1, we have raised the PostgreSQL requirements to version 12. And, while that should be a trivial change (and it is for GHA), for Travis they decided to change the default user and the default port for PostgreSQL >= 11.

Link: https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version

> For PostgreSQL 11 and 12 you must specify the packages, but the user is travis and the port is 5433 instead. So you must specify the PGPORT

I already have a PR near finished that, pseudo-automatically (based on an env variable called `PGVER`) decides between:
- db-user: postgres or travis.
- db-port: 5432 or 5433.

And the only piece that is missing... is exactly this PR, aka, add support for db-port (within db-options). Hence, it has my (self-interested) +1.

If you cannot do it, @geichelberger , would you mind if we add a commit to your PR branch amending the CHANGELOG, as @kabalin asked some days ago?

Ciao :-)

geichelberger:
Hi, I was busy the last few days and planned to do it today ;)

stronk7:
For the records, np about the master failures above (requiring PostgreSQL 12 for 4.1dev). That's exactly what my PR will be fixing, in the best possible way, trivial for GHA, but specifically for travis, where we'll be using this new db-port option to woraround the mess they did.

Ciao :-)

stronk7:
Sorry, forgot to say thanks, @geichelberger !

res/template/config.php.txt
```diff
@@ -11,7 +11,9 @@ $CFG->dbname    = '{{DBNAME}}';
 $CFG->dbuser    = '{{DBUSER}}';
 $CFG->dbpass    = '{{DBPASS}}';
 $CFG->prefix    = 'mdl_';
-$CFG->dboptions = [];
+$CFG->dboptions = [
+    'dbport' => '{{DBPORT}}',
```
scara, line 15: Hi @geichelberger and @stronk7,
should be `dbport` a string or an integer?

Moodle [uses](https://github.com/moodle/moodle/blob/d9632ca8b28304d5188a8310480f0cd6bc06d9d8/config-dist.php#L59-L61) a string, wondering if it is just for an "easy" fallback to the default port which in turns it's an actual empty string in the CLI command.

HTH,
Matteo

kabalin: I would say using string is OK here, as it does not seem important. It is responsibility of db engine class representing Moodle database interface to handle port param for connection: for MySQL it is [casted to int](https://github.com/moodle/moodle/blob/master/lib/dml/mysqli_native_moodle_database.php#L555), while for Postgres, it is [treated as string](https://github.com/moodle/moodle/blob/master/lib/dml/pgsql_native_moodle_database.php#L157).

kabalin (approved):
The change looks good to me.

One thing, rather than conditionally inject `--port` to connection string (in `src/Installer/Database/...`), we can have it permanently defined in the string and set specific to engine default port in the class variable along with type (which is then overridden if custom value is provided). Thinking more, in fact probably not the good idea in case if any of those clients can read port param from env variable, as having port always set via suggested change may break some installations...

Anyway, good to merge (if @scara is happy with my comment re type). Thanks @geichelberger!

stronk7:
Some 👍 here would be really appreciated, @scara (only if it's ok for you, of course!). This is stopping one more incoming PR (4.1 requirements) and a release to be done immediately after that.

TIA!

scara:
Hi @stronk7,
oops... My apologies, missed the question from @kabalin.

Here is my 👍 based on the comments above which makes me confident about the db-port-as-a-string usage.

HTH,
Matteo

stronk7:
Thanks @scara and @geichelberger !

I'm going to merge this now on behalf of @kabalin (thanks too!).

Ciao :-)

stronk7:
Merging this no matter Travis is failing with master branch. It's known issue (new Moodle 4.1 requirements) and will be fixed soon by another PR, based on this.

PR #165, opened 2022-04-21
PHPUnit should fail if there are warnings

marinaglancy:
When there are warnings in the phpunit results the exit code is 0.

The typical warning we receive is incorrect "@covers" tag. The pipeline does not fail but the output shows that there are warnings:

```
$ . check moodle-plugin-ci phpunit --coverage-text
 RUN  PHPUnit tests for tool_tenant
Moodle 4.0 (Build: 20220419), b10a543752f72d95987b8672ecf0a1348cdc9bac
Php: 7.3.33, pgsql: 10.20 (Debian 10.20-1.pgdg90+1), OS: Linux 4.4.0-1052-aws x86_64
PHPUnit 9.5.13 by Sebastian Bergmann and contributors.
..........S.S.......................................WWWWW......  63 / 191 ( 32%)
........................................SSSSSSS................ 126 / 191 ( 65%)
.....................................................WWWWWWWWWW 189 / 191 ( 98%)
WW                                                              191 / 191 (100%)
Time: 00:33.950, Memory: 319.00 MB
There were 17 warnings:
1) tool_tenant\external_dashboard_test::test_reset_linked_dashboard
"@covers \tool_tenant\external\dashboard\" is invalid
phpvfscomposer:///builds/workplace/moodle/vendor/phpunit/phpunit/phpunit:97
2) tool_tenant\external_dashboard_test::test_reset_tenant_dashboard
"@covers \tool_tenant\external\dashboard\" is invalid
phpvfscomposer:///builds/workplace/moodle/vendor/phpunit/phpunit/phpunit:97
...
17) tool_tenant\users_report_test::test_subtenant_users_own_sub_tenant_showall
"@covers \tool_tenant\users_report" is invalid
phpvfscomposer:///builds/workplace/moodle/vendor/phpunit/phpunit/phpunit:97
WARNINGS!
Tests: 191, Assertions: 1202, Warnings: 17, Skipped: 9.
Code Coverage Report:        
  2022-04-21 20:36:07        

```

To test the patch I created test plugin that has warnings in @covers directives, here is the GHA file: https://github.com/marinaglancy/moodle-tool_testpluginci/blob/master/.github/workflows/test.yml
and here is the result of the GHA: https://github.com/marinaglancy/moodle-tool_testpluginci/runs/6117617797?check_suite_focus=true

### DIFF
--- a/gha.dist.yml
+++ b/gha.dist.yml
@@ -103,7 +103,7 @@ jobs:
 
       - name: PHPUnit tests
         if: ${{ always() }}
-        run: moodle-plugin-ci phpunit
+        run: moodle-plugin-ci phpunit --fail-on-warning
 
       - name: Behat features
         if: ${{ always() }}
--- a/src/Command/PHPUnitCommand.php
+++ b/src/Command/PHPUnitCommand.php
@@ -30,7 +30,11 @@ class PHPUnitCommand extends AbstractMoodleCommand
         $this->setName('phpunit')
             ->setDescription('Run PHPUnit on a plugin')
             ->addOption('coverage-text', null, InputOption::VALUE_NONE, 'Generate and print code coverage report in text format')
-            ->addOption('coverage-clover', null, InputOption::VALUE_NONE, 'Generate code coverage report in Clover XML format');
+            ->addOption('coverage-clover', null, InputOption::VALUE_NONE, 'Generate code coverage report in Clover XML format')
+            ->addOption('fail-on-incomplete', null, InputOption::VALUE_NONE, 'Treat incomplete tests as failures')
+            ->addOption('fail-on-risky', null, InputOption::VALUE_NONE, 'Treat risky tests as failures')
+            ->addOption('fail-on-skipped', null, InputOption::VALUE_NONE, 'Treat skipped tests as failures')
+            ->addOption('fail-on-warning', null, InputOption::VALUE_NONE, 'Treat tests with warnings as failures');
     }
 
     protected function initialize(InputInterface $input, OutputInterface $output)
@@ -73,6 +77,11 @@ class PHPUnitCommand extends AbstractMoodleCommand
         }
         if ($this->supportsCoverage() && $input->getOption('coverage-clover')) {
             $options[] = sprintf('--coverage-clover %s/coverage.xml', getcwd());
+        }
+        foreach (['fail-on-warning', 'fail-on-risky', 'fail-on-skipped', 'fail-on-warning'] as $option) {
+            if ($input->getOption($option)) {
+                $options[] = '--'.$option;
+            }
         }
         if (is_file($this->plugin->directory.'/phpunit.xml')) {
             $options[] = sprintf('--configuration %s', $this->plugin->directory);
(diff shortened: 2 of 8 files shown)


gha.dist.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
       - name: PHPUnit tests
         if: ${{ always() }}
-        run: moodle-plugin-ci phpunit
+        run: moodle-plugin-ci phpunit --coverage-text --fail-on-warning
```
stronk7, line 106: To enable here in the dist/reference file `--coverage-text` ... I'm not sure if it's a good idea. Right now we are using `phpdbg` to generate code coverage and it's crashing badly (GHA only, AFAIK), specially with PHP 8.0 and up (I had to disable it in a number of repos, see for example https://github.com/moodlehq/moodle-local_codechecker/commit/8f8ca5d4b4bb241f18d765d145a129e630d3cf63).

So, until we move to pcov (that seems to work way more reliably), I'd suggest not to "promote" it in the dist file.

About having the `--fail-on-warning` proposed there, np, IMO.

src/Command/PHPUnitCommand.php
```diff
@@ -74,6 +78,11 @@ private function resolveOptions(InputInterface $input)
         if ($this->supportsCoverage() && $input->getOption('coverage-clover')) {
             $options[] = sprintf('--coverage-clover %s/coverage.xml', getcwd());
         }
+        foreach (['fail-on-warning', 'fail-on-risky', 'fail-on-skipped', 'fail-on-warning'] as $option) {
+            if ($input->getOption($option)) {
+                $options[] = '--'.$option;
```
stronk7, line 83: I bet this is the php-cs-fixer problem that is happening @ travis and GHA. This project doesn't use spaces around the concat operator. Sharing because it's tricky to find and it has happened to me in the past. ;-)

stronk7: Ah, you found it! LOL. Just refreshed!

stronk7:
Hi @marinaglancy,

would you mind applying this patch to your PR? That should solve travis problems, if I'm not wrong. Alternatively I think that I can push a commit to your branch (never have done that previously), if that's ok for you.

Ciao :-)

Here it's the patch, it's making self tests to pass here:

```
diff --git a/tests/Fixture/moodle-local_ci/tests/lib_test.php b/tests/Fixture/moodle-local_ci/tests/lib_test.php
index 287aa77..8143487 100644
--- a/tests/Fixture/moodle-local_ci/tests/lib_test.php
+++ b/tests/Fixture/moodle-local_ci/tests/lib_test.php
@@ -52,7 +52,7 @@ class lib_test extends \basic_testcase {
     /**
      * Test subtraction.
      *
-     * @covers ::local_ci_substract
+     * @covers ::local_ci_subtract
      */
     public function test_local_ci_subtract() {
         $this->assertEquals(0, local_ci_subtract(2, 2));
@@ -63,7 +63,7 @@ class lib_test extends \basic_testcase {
     /**
      * Test math class.
      *
-     * @covers \local_ci\math\add
+     * @covers \local_ci_math::add
      */
     public function test_local_ci_math() {
         $math = new \local_ci_math();
@@ -75,7 +75,7 @@ class lib_test extends \basic_testcase {
     /**
      * Test math class.
      *
-     * @covers \local_ci\math\add
+     * @covers \local_ci\math::add
      */
     public function test_local_ci_math_class() {
         $math = new math();
```

gha.dist.yml
marinaglancy: yes, makes sense. I've been struggling to run coverage tests on php8 in gitlab as well :(

stronk7 (approved):
Thanks Marina, I think it now looks perfect!

PR #167, opened 2022-05-10
Remove .phpcs.xml files immediately after creating it

stronk7:
moodle-plugin-ci is not ready for this yet. We need to move
from current use of local_codechecker to composer + moodle-cs
before being able to allow these new .phpcs.xml files in core.

So, for now, we are just deleting it as part of the tool
installation. This will be revisited when moving to the
new approach commented above.


stronk7:
Thanks @jrchamp,

I've applied your suggestions in new commit. Squashing them now...

stronk7:
Self-merged, towards making a release today.

PR #171, opened 2022-06-17
Move to moodle cs

stronk7:
In summary, this makes moodle-plugin-ci to stop using local_codechecker as a requirement and starts using the, now split, moodle-cs standard (with both it and PHP_CodeSniffer being installed via composer).

That was, any new sniff added to the standard and released, or any new PHP_Codesniffer version will be used immediately, without having to make "fake" local_codechecker releases anymore.

Note this PR has 2 commits:

1. The first just does the switch but continues using PHP_Codesniffer as a library, within PHP like it was an API. This API use was causing some problems, specially like #167, affecting all current versions (both before and after the commit).
2. The second commit makes the change from library to binary so, now, we use the `phpcs` binary passing everything from CLI / shell execution. It's way clearer and less hacky.3. 

All the existing features have been apparently (covered with new tests) preserved and all the testing, namely:
- moodle-plugin-ci itself
- local executions
- github, travis and gitlab executions

Are passing without the problems commented in #167 any more.

And that's this. One less tool using local_codechecker (the last one, so far), so we can continue with https://github.com/moodlehq/moodle-local_codechecker/pull/193 (making codechecker itself to use the standalone standard).

Ciao :-)

### DIFF
--- a/src/Command/CodeCheckerCommand.php
+++ b/src/Command/CodeCheckerCommand.php
@@ -28,7 +28,7 @@ use Symfony\Component\Finder\Finder;
 //
 // The alternative is to, instead of using PHP_CodeSniffer like a library, just
 // use the binaries bundled with it (phpcs, phpcbf...) but we want to use as lib for now.
-require_once __DIR__.'/../../vendor/moodlehq/moodle-local_codechecker/phpcs/autoload.php';
+require_once __DIR__.'/../../vendor/squizlabs/php_codesniffer/autoload.php';
 
 /**
  * Run Moodle Code Checker on a plugin.
@@ -110,7 +110,7 @@ class CodeCheckerCommand extends AbstractPluginCommand
         //
         // Note: the paths are relative to the base CodeSniffer directory, aka, the directory
         // where "src" sits.
-        $runner->config->setConfigData('installed_paths', './../PHPCompatibility');
+        $runner->config->setConfigData('installed_paths', './../../phpcompatibility/php-compatibility/PHPCompatibility');
         $runner->config->standards = [$this->standard]; // Also BEFORE init() or it's lost.
 
         $runner->init();
--- a/src/Command/CodeCheckerCommand.php
+++ b/src/Command/CodeCheckerCommand.php
@@ -27,9 +27,9 @@ class CodeCheckerCommand extends AbstractPluginCommand
     use ExecuteTrait;
 
     /**
-     * @var string Path to the temp file where the summary report results will be stored
+     * @var string Path to the temp file where the json report results will be stored
      */
-    protected $summaryFile;
+    protected $tempFile;
 
     protected function configure()
     {
@@ -46,7 +46,7 @@ class CodeCheckerCommand extends AbstractPluginCommand
     {
         parent::initialize($input, $output);
         $this->initializeExecute($output, $this->getHelper('process'));
-        $this->summaryFile = sys_get_temp_dir().'/moodle-plugin-ci-code-checker-summary-'.time();
+        $this->tempFile = sys_get_temp_dir().'/moodle-plugin-ci-code-checker-summary-'.time();
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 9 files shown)


src/Command/CodeCheckerCommand.php
```diff
@@ -63,106 +45,69 @@ protected function configure()
...
-                    return $e->getCode();
-                } catch (\Exception $e) {
-                    $error = 'Problem during processing; checking has been aborted. The error message was: '.$e->getMessage();
-                    $file->addErrorOnLine($error, 1, 'Internal.Exception');
-                }
-                $file->cleanUp();
-            }
+        // Add the files to process.
+        foreach ($files as $file) {
+            $builder->add($file);
```
kabalin, line 89: I suppose we need to respect `($file->ignored === false)` before adding file?

src/Command/CodeCheckerCommand.php
```diff
@@ -63,106 +45,69 @@ protected function configure()
...
+            ->setTimeout(null);
+
+        // If we aren't using the max-warnings option, then we can forget about warnings and tell phpcs
+        // to ignore them for exit-code purposes (still they will be reported in the output).
+        if ($input->getOption('max-warnings') < 0) {
+            $builder->add('--runtime-set')->add('ignore_warnings_on_exit')->add(' 1');
+        } else {
+            // If we are using the max-warnings option, we need the summary report somewhere to get
+            // the total number of errors and warnings from there.
+            $builder->add('--report-summary='.$this->summaryFile);
```
kabalin, line 84: Thinking if the better approach could be using `--report=json`, from which the number of warnings and errors can be retrieved without need of string parsing?

src/Command/CodeFixerCommand.php
```diff
@@ -46,90 +36,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            ->add($output->isDecorated() ? '--colors' : '--no-colors')
+            ->add('--report-full')
+            ->add('--report-width=132')
+            ->add('--encoding=utf-8')
+            ->setWorkingDirectory($this->plugin->directory)
+            ->setTimeout(null);
+
+        // Add the files to process.
+        foreach ($files as $file) {
+            $builder->add($file);
```
kabalin, line 62: I suppose we need to respect ($file->ignored === false) before adding file?

kabalin (commented):
Hi @stronk7 , thank  for working on this, great improvement. I left some comments for your consideration.

src/Command/CodeCheckerCommand.php
stronk7: Yeah, from codesniffer source, it seems that the `$file->ignore` is used internally when the whole file is going to be ignored:

```
      /**
       * If TRUE, the entire file is being ignored.
       *
       * @var boolean
       */
      public $ignored = false;
```

And I think it's used when there are embed tags (`// phpcs ignore`...) and things like that. I don't see any place where it's being used to apply any list of excludes. It's just set when parsing the file.

So, conclusion... I think we are filtering (excluding) the files already (thirdtparty libraries, ignore paths and ignore names constants...) and weren't using any PHP_CodeSniffer filtering/excluding ability at all. And we have changed that here.

Tangentially related... the only point I was a little bit worried is if, now that we are executing the binary (from shell), maybe, the list of files could become too big (shell command limits) and I did some quick tests and accepted a fairly large list without problems. So finally assumed that it was ok to pass the files as part of the command (it's also possible to pass the list in a file, just in case we face problems with command max length).

src/Command/CodeCheckerCommand.php
stronk7: Done, I've added a 3rd, fixup commit, with the switch from summary to json report format to get the total errors and warnings.

stronk7:
I'm resolving the existing conversations, think both are now ok:

1) The `->ignore` bit is ignorable (CodeSniffer internal stuff)
2) The count of errors and warnings now is done using the json report.

Once this is approved, I'll squash the `!fixup` commit(s) where correspond.

Ciao :-)

stronk7:
Merged, thanks @kabalin !

PR #173, opened 2022-06-28
Prepare 3 3 0

stronk7:
I've decided to bump it to 3.3.0 (instead of 3.2.6) because of the internal changes from moodle-local_codechecker to moodle-cs. Just in case anybody needs to staty sticky to 3.2.x or something like that.

Also, the CLI docs have been updated with auto-generated new options. And the version has been updated in the new place (bin/moodle-plugin-ci) as indicated by the release instructions.

Don't forget that this has to be added WITHOUT merge commit, use squash or rebase instead.

Ciao :-)


kabalin (approved):
Looks good to me!

Issue #151, opened 2022-01-12
Towards a new major version 4


stronk7:
Copied from a discussion elsewhere, just registering it here for further comments and planning:

> ...
> Maybe:
> 
> 1. We should consider to start working in a new major (4.0.0) release, coming with:
>     - move php mis requirement to php 7.3 (or 7.4). That will allow people to continue testing Moodle 3.6.4 ( or 3.8.3) and up (for older versions the 3.x series still will be working).
>     - updated components (symfony and friends).
>     - move to pcov for coverage maybe
>     - update phpunit/self tests…
> 2. Still, not matter the work in master towards that 4.0.0 new version, if we need to update anything in stables… we always can create a “3.x” branch and apply incremental changes there.
> 
> My main point is to move away from current php 7.0 compatibility (Moodle 3.0.1) that is forcing us to keep other tools (codechecker, symfony…) using really old stuff and, at some point, we’ll face real problems (for example with php81).

stronk7:
Just experimenting:

[![](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.2%2C%20supporting%20Moodle%203.4%20and%20later%20releases)](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.2%2C%20supporting%20Moodle%203.4%20and%20later%20releases/vote)
[![](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.3%2C%20supporting%20Moodle%203.6.4%20and%20later%20releases)](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.3%2C%20supporting%20Moodle%203.6.4%20and%20later%20releases/vote)
[![](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.4%2C%20supporting%20Moodle%203.8.3%20and%20later%20releases)](https://api.gh-polls.com/poll/01FS7K69XDZ4817NJ7RBD82SZ2/PHP%207.4%2C%20supporting%20Moodle%203.8.3%20and%20later%20releases/vote)
[...]

PR #176, opened 2022-07-07
Support and document usage with the Moodle App

NoelDeMartin:
I've been trying to use this for a plugin with mobile support, and fortunately it seems like very few changes are necessary to make it work.

The most tricky is changing the default browser though. I guess it would be better to just use chrome as the default, but this could potentially break some existing tests so I just decided to keep the same behaviour by default.

Other than a few changes in the code itself, I've documented the configuration. [I've managed to get this working in Github Actions](https://github.com/NoelDeMartin/moodle-local_sandbox/actions/runs/2629805991), but I wasn't able to test the Travis config because I don't have any more credits in my account :sweat_smile: I think it should work though, because the changes are minimal and very similar to the ones done for Github Actions.


jrchamp (commented):
Just a small note (not a full review), there's two references in here to postgres 10 that should probably be updated to postgres 12 to match the recent version bump.

NoelDeMartin:
Hey @jrchamp, thanks for bringing that up. I initially implemented this before postgres 12 was used, and I didn't notice that this had changed so I didn't update it when rebasing. It's fixed now.

stronk7 (changes requested):
Hi @NoelDeMartin ,

I like this change. Apart from the two suggestions that I've added as comments in the review... surely the unique point that I'm missing here is to have some own CI verifying that the new app support is working as expected. Maybe we can do something like:
- Add some "easy" mobile test to the fake plugin that we use for self-testing (`tests/Fixture/moodle-local_ci`).
- Ensure that both our Travis and GHA tests run it once? I'd suggest, for now, to add them just to the "MOODLE_401_STABLE" branch (add the 2 extra env variables there) and try it.

I've not tried, but I think that it will work and, that way, we ensure that we are always at very least running the apps stuff once.

What do you think?

Ciao :-)

NoelDeMartin:
Thanks @stronk7 for your review! I've applied your suggestions and tweaked a couple of things to add tests, so please take another look.

Something I've had to do is removing the tests with `MOODLE_38_STABLE`, because the older behat configuration in the LMS is not compatible with the current version of the app. But I think it should be ok to remove those, given that 3.8 is already deprecated.

stronk7:
Hi @NoelDeMartin,

A) before everything else... I want to share that, right now, we are (in the `master` branch) preparing the next big release (4.0.0) of moodle-plugin-ci. And a lot of changes have been applied to it already, with that, surely, being the cause of the conflicts that this PR is showing right now.

Also, note that there is the `3.x` branch for any stuff that we want to get "backported" there.

For more information, #151 is where all the `v4` progress is being shared.

Said that... maybe it wouldn't be a bad idea to make this one of those `v4` features, so we keep `3.x` stable and we don't need to go up-porting big things to `master`.

What do you think?

B) Second... one thing that I'm not sure I like (from a developer perspective) is the need to add any new step to install the `moodle-local_moodleappbehat` plugin and to start the `moodlehq/moodleapp`docker container. So I was guessing if, maybe... it could have sense to create a new command, say, `behatapp` that is able to, behind the scenes... do those things (install the local plugin and start the docker container).

Note that we want to do something similar with the phpunit and behat commands (not done yet, see #176), so the databases (for example) are only installed if the corresponding command is used (so the proposed change here for that new `behatapp` command would match nicely that behaviour (only install requirements if the command is effectively used).

What do you think?

C) Apart from the above points (or maybe related, heh), I'm not sure why we need specialised "dist" files. If, after all, it's the command (`behatapp`) the one going to decide what's instantiated and executed... surely the very same "dist" files will server for all commands. Or, in any case, no matter if B) is implemented or no... still... I think that it's more clear if we just keep one dist file, with commented-out stuff and/or more detailed comments in the "Explained" docs.

D) Also, there are a few details... in the proposed patch that I'm not sure to fully understand, like a) why do we need to specify the plugin in the self-testing, b) the lack of travis self-testing... note that the PRs DO run the travis self-tests, so we can ensure the app is working there too,  c) the removal of 38_STABLE from the self-tests (it continues being supported officially in `v4`, d) changing the default browser... and surely some other small detail.

In any case... the main two points here are, IMO, A) and B). Aka:

A) Which the target of this PR is going to be. v3 and v4 or only v4.
B) If we try to embed all the dependencies within a new `behatapp` command, instead of having to install/launch them manually in the CI scripts.

Once those 2 are agreed, I think we can start working on everything with clearer goals.

Ciao :-)

NoelDeMartin:
Thanks for checking it out @stronk7,

A) I didn't know this was happening, but yes I agree with adding it only for 4.0.

B) I also agree the DX would be better like that, but I didn't want to do so many changes (although I think at this point I've already changed a lot xD). What I don't like is using a separate command. Maybe we could do something like adding an `--app` flag or something in the `install` command? And then just use the normal `behat` command to run the tests.

C) Ok, if we end up removing all those special steps I think it makes sense to keep only one dist file.

D.a) I needed to do that because after installing `moodlehq/moodle-local_moodleappbehat`, the commands didn't work without specifying the plugin. I'm not sure if that's the expected behaviour or I was doing something wrong. You can see the error I got in CI here: https://github.com/NoelDeMartin/moodle-plugin-ci/actions/runs/3638952429/jobs/6141768058

D.b) I mentioned it in the description, but I can't test with Travis because I don't have any credits in my account :/ I guess we can leave this for the end and when everything else is working, I can ask someone for help or use one of the paid accounts we have for the apps team.

D.c) I mentioned this in my last comment, but LMS 3.8 is not compatible with the current architecture to execute app tests. I thought it was fine to remove since 3.8 is not a supported version anymore, but if that's a problem I'll try to configure it so that the tests are executed skipping the app tests.

NoelDeMartin:
@stronk I've updated the code with the latest version, so you should be able to take a look again.

Besides applying what we discussed last time, I've also simplified the setup. For the most part, it should work just by setting a new `MOODLE_APP` env variable which is opt-in. If that variable is missing, everything should work the same way as before the PR.

Other than that, I also dropped the tests with the 3.10 version because they were [giving some failures](https://github.com/NoelDeMartin/moodle-plugin-ci/actions/runs/6955796196/job/18925342091) that I couldn't sort out. They were not related to the app either, and given that 3.10 is not a supported version anymore, it should be fine to ignore.

Everything else should be explained in the documentation, let me know if there is something else I should change.

.github/workflows/test.yml
```diff
@@ -145,6 +142,7 @@ jobs:
         moodle-plugin-ci phpunit --verbose --coverage-text --fail-on-warning
         moodle-plugin-ci behat --profile default
         moodle-plugin-ci behat --profile chrome
+        moodle-plugin-ci behat --profile firefox --tags="@local_ci&&~@app"
```
stronk7, line 145: I'm a little bit confused with this. How we are specifying firefox and get the `@app` tests run?

More yet... as far as me have set `MOODLE_APP: 'true'` on install... doesn't that mean that the 2 behat executions above (default and chrome) will be also running the `@app` tests (because the new added feature has `@local_ci` it will be run by default.

Not sure if I'm explaining, heh. Apart from the initial question, it's not clear for me why to specify `@app` is needed. I see it if we only want to run `@app` tests but not if we want to run all the `@local_ci` tests. And, in any case, how it's possible that they are passing with `--profile default` or `--profile firefox` profile?

.github/workflows/test.yml
```diff
@@ -258,3 +257,4 @@ jobs:
         php build/moodle-plugin-ci.phar phpunit --verbose --coverage-text --fail-on-warning
         php build/moodle-plugin-ci.phar behat --profile default
         php build/moodle-plugin-ci.phar behat --profile chrome
+        php build/moodle-plugin-ci.phar behat --profile firefox --tags="@local_ci&&~@app"
```
stronk7, line 260: Same than above... I imagine I'll understand later when arriving to the code, but right now it doesn't fit in my brain, hence annotating.

.travis.yml
```diff
@@ -46,7 +46,7 @@ before_install:
     fi
 
 install:
-  - moodle-plugin-ci install -vvv
+  - MOODLE_APP=true moodle-plugin-ci install -vvv
```
stronk7, line 49: While this is correct, maybe I'll set the variable above (`global.env`), just to be like the rest of main variables and because the dist and explained files are putting it there.

.travis.yml
```diff
@@ -63,6 +63,7 @@ script:
   - moodle-plugin-ci phpunit --verbose --coverage-text --fail-on-warning
   - moodle-plugin-ci behat --profile default
   - moodle-plugin-ci behat --profile chrome
+  - moodle-plugin-ci behat --profile firefox --tags="@local_ci&&~@app"
```
stronk7, line 66: Heh, again. Annotated.

.github/workflows/test.yml
```diff
@@ -86,12 +86,8 @@ jobs:
             moodle-branch: 'MOODLE_400_STABLE'
           - php: '8.0'
             moodle-branch: 'MOODLE_311_STABLE'
-          - php: '7.4'
-            moodle-branch: 'MOODLE_310_STABLE'
           - php: '7.4'
             moodle-branch: 'MOODLE_39_STABLE'
-          - php: '7.4'
-            moodle-branch: 'MOODLE_38_STABLE'
```
stronk7, line 94: Regarding removing 38 and 310 completely... I must confess that I'm not 100% convinced... maybe we could add some variable, say, `moodle-app`(o `skip-moodle-app`) to the supported branches in the matrix and then set the  `MOODLE_APP` based on that?

That way we could keep the feature enabled and disabled at our preference, not only to disable it in the 38 and 310 branches (obviously), but to:
1. Continue having 38 and 310 covered for everything else.
2. Maybe split some run (main branch) and have it both with the app enabled and disabled (just to verify that everything continues working with the app disabled).
3. Slightly better real coverage (I'll comment later about better phpunit coverage).

.travis.yml
```diff
@@ -94,6 +95,6 @@ jobs:
     # Last stable with highest supported PHP version.
     - php: 8.2
       env: MOODLE_BRANCH=MOODLE_402_STABLE
-    # And oldest stable supported (with lowest supported PHP version).
+    # And older stable supported (with lowest supported PHP version).
     - php: 7.4
-      env: MOODLE_BRANCH=MOODLE_38_STABLE
+      env: MOODLE_BRANCH=MOODLE_39_STABLE
```
stronk7, line 100: Same comment here than in the GH workflow self-tests. Maybe we can add that extra setting to decide...

docs/MoodleApp.md
```diff
@@ -0,0 +1,20 @@
...
+title: Moodle App
+---
+
+In order to test plugins with mobile support, the only command that needs a special configuration is `behat`.
+
+In practice, you only need to set the `MOODLE_APP` env variable to `true`, and all the dependencies will be configured during the `install` command. You can also configure the behaviour of [Acceptance testing for the Moodle App](https://moodledev.io/general/app/development/testing/acceptance-testing) using the following env variables:
+
+- `MOODLE_APP_DOCKER_IMAGE`: Tag of [the Moodle App Docker image](https://moodledev.io/general/app/development/setup/docker-images) to use for running the app. The default value is `moodlehq/moodleapp:latest-test`.
+- `MOODLE_APP_BEHAT_PLUGIN_PROJECT`: Project in github to use for installing the plugin with Behat steps specific to the Moodle App. The default value is `moodlehq/moodle-local_moodleappbehat`. This variable will be ignored if `MOODLE_APP_BEHAT_PLUGIN_REPOSITORY` is set.
+- `MOODLE_APP_BEHAT_PLUGIN_REPOSITORY`: Repository url to use for installing the plugin with Behat steps specific to the Moodle App. By default, the github repository defined in `MOODLE_APP_BEHAT_PLUGIN_PROJECT` will be used.
```
stronk7, line 12: Why do we need this duality? Sure there is a reason, but I cannot get it now.

docs/MoodleApp.md
```diff
@@ -0,0 +1,20 @@
...
+- `MOODLE_APP_BEHAT_PLUGIN_BRANCH`: Branch of the repository to use for installing the plugin with Behat steps specific to the Moodle App. The default value is `latest`.
```
stronk7, line 13: Does this apply both to `MOODLE_APP_BEHAT_PLUGIN_PROJECT` and `MOODLE_APP_BEHAT_PLUGIN_REPOSITORY` ?

docs/MoodleApp.md
```diff
@@ -0,0 +1,20 @@
...
+- `MOODLE_BEHAT_IONIC_WWWROOT`: Value to use in `$CFG->behat_ionic_wwwroot`. The default value is `http://localhost:8100`. This value should only be used if `MOODLE_APP` is not set, and the Moodle App dependencies are configured manually.
+
+Finally, keep in mind that mobile tests only run on chrome device, so make sure to use it with the `--profile` flag.
```
stronk7, line 16: Given how we are using in the self-tests... and how the template and explained files are... is this paragraph correct?

tests/Fake/Process/DummyExecute.php
```diff
@@ -18,6 +18,7 @@
 class DummyExecute extends Execute
 {
     public string $returnOutput = '';
+    public array $cmds          = [];
```
stronk7, line 21: Maybe `$allCmds`, silly detail, but...

tests/Installer/MoodleAppInstallerTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace MoodlePluginCI\Tests\Installer;
+
+use MoodlePluginCI\Installer\MoodleAppInstaller;
+use MoodlePluginCI\Tests\Fake\Process\DummyExecute;
+use MoodlePluginCI\Tests\MoodleTestCase;
+
+class MoodleAppInstallerTest extends MoodleTestCase
+{
+    public function testInstall()
```
stronk7, line 21: Maybe here we could add some more tests, setting up different configuration options (MOODLE_APP, other globals...). Just towards getting some of those logical branches in the AppInstaller covered.

I also would add some tests to the BehatCommand tests to verify that the new options are processed ok (when defined).

stronk7 (changes requested):
Hi Noel,

thanks for the great stuff and the patience! I've added a bunch of review-comments here and there, let's go discussing them.

Tomorrow I'll be doing some local tests (phar integration tests...) and surely will add some comment with a few points with whatever I find.

I'm sure that in a couple of iterations we can have this ready to become onboard!

Ciao :-)

.github/workflows/test.yml
NoelDeMartin: The run with firefox is doing precisely the opposite, running all the tests except the app's (notice the `~` before the `@app` tag, that means to skip the tests with the `@app` tag).

I added this line in the CI because when `MOODLE_APP` is set to true, the default browser is chrome. So without this, there wouldn't be any tests using firefox in the CI and we still want to make sure that non-app tests continue working.

docs/MoodleApp.md
NoelDeMartin: The first one, `MOODLE_APP_BEHAT_PLUGIN_PROJECT`, is used if the plugin is hosted on github so you can just set it as `moodlehq/moodle-local_moodleappbehat`; whilst the second one, `MOODLE_APP_BEHAT_PLUGIN_REPOSITORY`, takes a full url so you can use repositories that aren't hosted in github, like `https://gitlab.com/moodlehq/moodle-local_moodleappbehat.git` (this url doesn't exist, just illustrating the idea).

This is actually the same that happens with the `--project` and `--clone` arguments in [the add-plugin command](https://github.com/moodlehq/moodle-plugin-ci/blob/main/src/Command/AddPluginCommand.php), that's why I did it. I guess I can change the naming or documentation if you want. But I think "repository" is a better name than "clone" for an env variable.

docs/MoodleApp.md
NoelDeMartin: Yes.

.travis.yml
NoelDeMartin: Answered above.

docs/MoodleApp.md
NoelDeMartin: You're right, I've rewritten this paragraph.

.github/workflows/test.yml
stronk7: Crap, believe me I checked everything to see if, maybe, `MOODLE_APP` was getting precedence over `--profile`... and I did not see the `~`. Now it's clear, resolving this, thanks!

docs/MoodleApp.md
stronk7: Ok, clear. Resolving.

stronk7: Maybe we can consider renaming the `--clone` option (while keeping it working for some good time) and use `--repository` there. I've created #260 about it.

.github/workflows/test.yml
stronk7: I will try to give to this a try myself on top of your branch. While it's not super-critical... I feel that it's like abandoning those 2 branches completely and prefer to keep them covered, no matter that doesn't include the new app functionality.

So, ignore this, I'll take a look. Resolved.

tests/Installer/MoodleAppInstallerTest.php
stronk7: I'm ok. We can always add missing coverage of logical branches later. Resolved.

.travis.yml
stronk7: I see, ok. NVM.

docs/MoodleApp.md
```diff
@@ -0,0 +1,20 @@
...
+- `MOODLE_APP_BEHAT_PLUGIN_BRANCH`: Branch of the repository to use for installing the plugin with Behat steps specific to the Moodle App. The default value is `latest`.
+- `MOODLE_BEHAT_IONIC_WWWROOT`: Value to use in `$CFG->behat_ionic_wwwroot`. The default value is `http://localhost:8100`. This value should only be used if `MOODLE_APP` is not set, and the Moodle App dependencies are configured manually.
+
+Finally, keep in mind that mobile tests only run on Chrome, so make sure that you're using the correct browser. If you're setting the `MOODLE_APP` variable, the default browser should already be Chrome; but you can set it explicitly using the `--profile` flag otherwise.
+
+For specific examples, look at the [gha.dist.yml](GHAFileExplained.md) or [.travis.dist.yml](TravisFileExplained.md) files, and uncomment the lines mentioning the app.
+
+**Important:** Please notice that `MOODLE_APP` only works starting with version 4.1 of the app. If you want to use this setup against older versions, you'll have to use `MOODLE_BEHAT_IONIC_WWWROOT` and configure the dependencies manually (installing the plugin, launching the docker image, etc.).
```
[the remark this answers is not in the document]
stronk7: A: Forget my question, it's `MOODLE_APP_DOCKER_IMAGE` the one in charge of deciding the app version to use, defaulting to `latest-test`. Resolving this now.

stronk7:
And, yay, it landed, well done!

As said, I'll try to "revive" the 38 and 310 self-tests in another issue but, other than that, this is done.

Ciao :-)

NoelDeMartin:
Thanks for integrating it @stronk7 :D.

About your question regarding old versions of the app, I don't think it's worth the effort because the app's lifecycle is not as long-lived as the LMS. When a new version is released, the old one is deprecated instantly and if we have to fix any issues we do it for the next release. That's why it only makes sense to test against the stable and development versions of the app.

Even so, if someone has a custom app and hasn't updated it for a while, they should still be able to make use of this by changing the default variables as written in the documentation.

And if you were mentioning it to configure some CI tests for multiple versions of the app... I don't think it's a good idea to test against the development version, because it can often be broken for known issues (right now for example we're migrating to a new version of Ionic and it'll be broken for a while). So I think it only makes sense to test it against the stable version, which is how it's already configured by default.

PR #177, opened 2022-07-27
Bump self-tests to 4.1 requirements

stronk7:
Trivial for GHA, just change of images to postgres:12, but...

Travis changed the PostgreSQL package for version 11 and up so, instead of
using the "postgres" user it now uses the "travis" one. And the port is
5433 instead of 5432. We use the existence of the PGVER environmental
variable to decide which defaults to use.

More yet, the connection via "localhost" (local net) now requires login and
password, it used to be trust/peer auth mode (not requiring password). If we want to
keep localhost (+ port) working, then we need to edit the  pg_hba.conf file
to trust/peer the local connections and then restart the database.

So, at the end, we are going to use socket connections (host = '')
that is perfectly ok for Travis (non dockered database). Only if they
haven't been configured another way manually (user, host, port).

Plus documenting the changes in a 2nd commit.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,16 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 ### Added
 - `moodle-plugin-ci install` now provides an option `--db-port` to define a custom database port.
 
+### Changed
+- Updated [gha.dist.yml](https://github.com/moodlehq/moodle-plugin-ci/blob/master/gha.dist.yml) and
+  [.travis.dist.yml](https://github.com/moodlehq/moodle-plugin-ci/blob/master/.travis.dist.yml)
+  (and documentation) to fulfil [Moodle 4.1 new requirements](https://tracker.moodle.org/browse/MDL-71747).
+- ACTION REQUIRED: Review existing integrations running tests against master (4.1dev). There are a few Moodle 4.1 new requirements:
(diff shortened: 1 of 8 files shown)


docs/CHANGELOG.md
```diff
@@ -12,6 +12,16 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 ### Added
 - `moodle-plugin-ci install` now provides an option `--db-port` to define a custom database port.
 
+### Changed
+- Updated [gha.dist.yml](https://github.com/moodlehq/moodle-plugin-ci/blob/master/gha.dist.yml) and
+  [.travis.dist.yml](https://github.com/moodlehq/moodle-plugin-ci/blob/master/.travis.dist.yml)
+  (and documentation) to fulfill [Moodle 4.1 new requirements](https://tracker.moodle.org/browse/MDL-71747).
```
paulholden, line 18: Typo: "fulfil" is the standard British/Australian spelling (although I admin I don't know which standard you are following already, so please feel free to ignore)

stronk7: Yep, British we are. I really didn't know that both singles existed, amending them now, thanks! Amended!

paulholden (approved):
Thanks @stronk7! Nice getting everything prepped for 4.1, approved

stronk7:
Thanks @paulholden , merged!

PR #178, opened 2022-07-27
Update tools to current ones

stronk7:
Not much to say, just ensure we are running the latest:
- local_ci
- local_moodlecheck


stronk7:
Self-merging this, towards an imminent release...

PR #183, opened 2022-09-12
Bump GHA environments to use Ubuntu 22.04

stronk7:
Since some time ago, GH is warning us that the Ubuntu 18.04 environment is deprecated:

"The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead...."

Hence, we are bumping here, to be able to cerify if everything continues working ok.

The changes cover own scripts, templates and guides. All them now pointing to Ubuntu 22.04 (jammy).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
     # Avoid forks to perform this job.
     if: github.repository_owner == 'moodlehq'
     name: Create Release
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Check out repository code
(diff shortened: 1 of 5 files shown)


kabalin (dismissed):
Thanks @stronk7 , we are future-proofed.

stronk7:
Note I'm removing the "Travis" part from this issue. They just rolled Ubuntu 22.04 (jammy) support days ago and I was having problems both with supported PostgeSQL and PHP versions.

So, finally, leaving the Travis side untouched, this is only GHA and I'm going to change the title now.

Ciao :-)

kabalin (approved):
Looks good!

stronk7:
Thanks, @kabalin !

jrchamp:
In case you want something more dynamic, `ubuntu-latest` has worked well for me.

stronk7:
Yeah,

but I prefer to avoid dynamic versions, they use to come with surprises here and there. Thanks!

PR #201, opened 2023-01-23
Update tools to current ones

stronk7:
- local_ci
- local_moodlecheck

Not much to say, just updating before making a quick release. The updates will solve a few composer warnings, espcially when installing the local_ci requirements, because there was a wrongly named file leading to this warning:

```
Class MoodleHQ\MoodleCS\moodle\Tests\MoodleCsStandardTest located in ./vendor/moodlehq/moodle-cs/moodle/Tests/MoodleStandardTest.php does not comply with psr-4 autoloading standard. Skipping.
```

Not the end of the world, because that Test file is not used (comes from moodle-cs), but towards a clean composer install.

Ciao :-)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,11 +40,11 @@
       "type": "package",
       "package": {
         "name": "moodlehq/moodle-local_ci",
-        "version": "1.0.15",
+        "version": "1.0.16",
         "source": {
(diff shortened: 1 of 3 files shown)


paulholden (approved):
LGTM

PR #207, opened 2023-03-04
Bump moodle-local_moodlecheck to v1.1.4

stronk7:
Also, update moodle-cs base version (though that
one is automatic on composer install/project).

And fix a few fixtures to pass with new phpcs
and node 18 stuff. Note this causes tests for 3.x
branches to fail with grunt, because the generated map files are now different from the 4.x ones. Hence, we are preventing that with conditional regular
expression for all 3.x branches.


stronk7:
I'm self-merging this, towards a small 3.x release, so I can up-port this to master and continue with 4.x stuff there.

PR #209, opened 2023-03-05
Bump all travis files and docs to require Ubuntu 20.04

stronk7:
NodeJS 18 requires GLIBC 2.28, that only comes in 18.10 (non LTS) and 20.04 (focal) Ubuntu versions. Change all the Travis references to point to that, because, by default, it uses denial older version.

Also bump the "make validate" self tests and the Travis dist files and docs to use php74 as minimum,,
because focal on Travis doesn't support php73 any more.

Finally, it seems that, for focal, they have changed again the PostgreSQL port to be 5432 (it was 5433 for PostgreSQL 11 and 12). So let's specify it again. Ping, pong!

Side note: We have disabled grunt for all Moodle 3.x branches because the Node 18 issue came with unexpected changes to .map files and now they are failing. [MDL-77527](https://tracker.moodle.org/browse/MDL-77527) has been created to fix that regression of [MDL-75012](https://tracker.moodle.org/browse/MDL-75012). Once it's fixed, we'll enable back grunt for 39 and 311.

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: focal
+
 addons:
   postgresql: "13"
   apt:
@@ -17,9 +19,9 @@ cache:
     - $HOME/.npm
 
 php:
- - 7.3
  - 7.4
  - 8.0
+ - 8.1
 
 env:
  global:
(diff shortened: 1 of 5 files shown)


paulholden (approved):
Thanks, I had the same problem locally with Ubuntu LTS

PR #210, opened 2023-03-06
Automatically add the version changelog section to the release notes

stronk7:
Using the yashanand1910/standard-release-notes action, now we are extracting the matching version from the docs/CHANGELOG.md file and adding it to the release notes.

This is an example of a released version (99.99.99) and how the release info looks:

https://github.com/stronk7/moodle-plugin-ci/releases/tag/99.99.99

And, this an example of a failed run because the released version (99.99.98) does not have any entry in the changelog:

https://github.com/stronk7/moodle-plugin-ci/actions/runs/4347058191

Ciao :-)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,11 +26,20 @@ jobs:
           make build
           php build/moodle-plugin-ci.phar list
 
+      - name: Extract release notes from changelog
+        id: extract-release-notes
+        uses: yashanand1910/standard-release-notes@v1.3.0
+        with:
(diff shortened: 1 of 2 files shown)


stronk7:
Hi, I'm going ahead with this because it has been tested (both here and in 3.x) and have another PR just merged and want to do a new release today / tomorrow. So perfect to have this available and verify it works as expected then.

I'll wait a couple of hours before proceeding, just in case anybody wants or has anything to object.

Ciao :-)

PR #212, opened 2023-03-08
CI Example Updates

### LINKED ISSUE #211: Clarify Code Coverage in CI Examples
> I think that `none` is the default int the dist files (both [GHA](https://github.com/moodlehq/moodle-plugin-ci/blob/master/gha.dist.yml#L49-L50) and [Travis](https://github.com/moodlehq/moodle-plugin-ci/blob/master/.travis.dist.yml#L37-L40)) because those dist files don't use any phpunit coverage option.
> 
> Only if some phpunit coverage option is used is when the `pcov => xdebug => phpdbg` fallback happens and they are used.
> 
> Surely we could:
> 1. Make the GHA dist file clearer (I think the Travis one is better).
> 2. At some point, go deprecating the `phpdbg` option. Maybe warning... that is a can of worms, specially with PHP 8.x versions.

_Originally posted by @stronk7 in https://github.com/moodlehq/moodle-plugin-ci/issues/204#issuecomment-1460700852_

jrchamp:
These comment changes include the additional context that I was previously missing to understand how code coverage works in moodle-plugin-ci.

Also, switches the examples to use `phpcs` instead of the `codechecker` alias.

Fixes #211 (deprecating phpdbg should probably be a separate issue)

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -34,9 +34,10 @@ env:
 before_install:
   - if [[ ${TRAVIS_PHP_VERSION:0:1} -gt 7 ]]; then pecl install xmlrpc-beta; fi
   - echo 'max_input_vars=5000' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
-  # Don't disable xdebug if you want to get it used for code coverage.
+  # We remove xdebug by default for performance, because we're not checking code coverage.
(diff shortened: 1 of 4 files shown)


stronk7 (approved):
Thanks @jrchamp !

I've created #213 about the deprecation / removal of `phpdbg`.

What else can I say, perfect patch, with nice wording (I'd have needed 10 paragraphs to explain the same, LOL).

Issue #146, opened 2021-12-23
Stop generating default coverage for master branch


andrewnicols:
Following MDL-72701 we should stop generating a default coverage file as this is we now generate a sensible configuration which covers classes, lib.php, locallib.php, and other relevant locations.

stronk7:
Closed by #216

PR #216, opened 2023-03-21
Don't overwrite the phpunit.xml file so much

### LINKED ISSUE #203: install overrides phpunit.xml file in project
When running moodle-plugin-ci (as it is the default in the provided github actions template) moodle-plugin-ci will create a new phpunit.xml file in the plugin directory, ignoring the one in my git repository. In my opinion this is counter-intuitive, because that way the pipeline is doing things differently than it is in every other environment and compared to how i configured it. 

In my case I have some files in the repo just for reference and some files containing common code in tests directory.  Both are included in the test coverage because of that issue. Also other options like `beStrictAboutOutputDuringTests` are different in the CI pipeline.

There is no option to disable only the phpunit.xml generation. As a workaround I'm currently using `moodle-plugin-ci install --no-init` and running `php admin/tool/phpunit/cli/init.php` manually.

**What i expected to happen**
- Using the phpunit.xml file in the repo (maybe with an exception if it differs from what is currently expected)
- Be clear about that issue in the documentation/logs
- Having a switch to disable phpunit.xml generation

**References**
- [Pipeline with the wrong phpunit.xml printed in step PHPUnit tests](https://github.com/ProjektAdLer/MoodlePlugin/actions/runs/4245418989/jobs/7380898537)
- [phpunit.xml in my repository](https://github.com/ProjektAdLer/MoodlePlugin/blob/927dc3a3ef28e0f6af39dd9585ecaa68e8081c46/phpunit.xml)
- [Workflow with workaround](https://github.com/ProjektAdLer/MoodlePlugin/blob/927dc3a3ef28e0f6af39dd9585ecaa68e8081c46/.github/workflows/moodle-ci.yml)

stronk7:
Fixes #203 and #146

This solves two cases where moodle-plugin-ci was being "too aggressive" managing the phpunit.xml files.

1. If the plugin comes with its own phpunit.xml file, give to it absolute precedence, so it's not overwritten or modified ever.
2. If the plugin doesn't have any tests/coverage.php file, instead of applying moodle-plugin-ci defaults, let's allow core to apply its own ones. See MDL-72701 for more info about core coverage defaults. We'll continue applying our defaults for Moodle < 3.9.


stronk7:
(if this is accepted and merged, I'll proceed to backport it to the 3.x branch, so a new release can happen there)

src/Bridge/MoodlePlugin.php
```diff
@@ -197,6 +197,18 @@ public function hasFilesWithName($pattern): bool
...
     }
 
+    /**
+     * Determine if the plugin has a (phpunit.xml) configuration file.
+     *
+     * @return bool
+     */
+    public function hasPhpUnitConfig(): bool
+    {
+        $result = $this->getFiles(Finder::create()->depth(0)->name('phpunit.xml'));
```
kabalin, line 207: I guess phpunit can be in any directory (otherwise it might be enough to do `file_exists($this->pluginDir . '/phpunit.xml')`)?

kabalin (dismissed):

src/Bridge/MoodlePlugin.php
stronk7: Yeah, I initially was thinking about to allow them in any subdir, but then realised that we only use those `phpunit.xml` files when they are on plugin base/root dir. And I never have seen them elsewhere.

https://github.com/moodlehq/moodle-plugin-ci/blob/master/src/Command/PHPUnitCommand.php#L105-L115

So I think it's ok to prevent recursion, we only care about the base/root one, that in fact is the one created by our `admin/tool/phpunit/cli/init.php --buildcomponentconfigs` option (the one we are preventing here to be used if there is already a `phpunit.xml` file created by the dev.

stronk7: About using simpler `file_exists()` vs `Finder` component, I saw that there were uses of both in the class and decided to go to the later, just to know/experiment more about it. There isn't much else to say. I'm happy either way.

kabalin: I suggest to use `file_exists` for better readability.

stronk7: Have amended it to use simpler `is_file($this->directory . '/phpunit.xml')`

PR #220, opened 2023-03-22
Add support for testdox output

### LINKED ISSUE #219: Add testdox option to phpunit command
testdox formatter generates nice output so it might be useful to have it.

kiklop74:
Fixes #219


stronk7:
Hi Darko,

have you tried this? I don't think that simply adding the option makes it to work. I think that, some lines later, in `resolveOptions()` you've to, also, add explicit support for it.

kiklop74:
Updated the code.

stronk7 (approved):
Thanks @kiklop74 ,

this looks good enough and coverage is not much affected, no matter the new option is not covered explicitly.

So, all ok, so far, thanks!

PR #222, opened 2023-03-27
Just give GHA a little bit priority over Travis

stronk7:
As far as we are removing Travis CI from core (see MDLSITE-7135) this just applies small tweaks to all the places where stuff related with Travis is documented, always trying to put the GHA alternative earlier in the doc.

Basically moving a few paragraphs around and some little details here and there.

I will backport this to the 3.x branch once/if accepted upstream.

Ciao :-)


kabalin (approved):
LGTM

PR #236, opened 2023-09-07
Update to current box version (phar packaging)

### LINKED ISSUE #235: Running plugin-ci on PHP 8.1.22 shows warnings
Hi,

I'm getting the following when trying to use the moodle-plugin-ci.phar downloaded from the latest released version v4.1.2

`Return type of Symfony\Component\Console\Helper\HelperSet::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///Users/jonathan.garcia/Downloads/moodle-plugin-ci-2.phar/vendor/symfony/console/Helper/HelperSet.php on line 104`

Found this while trying to run it on Linux (happens also in MacOs) for PHP 8.1.

This error also appears when running any command like "save points" or "grunt".

Any hep will very appreciated.

stronk7:
Current version (2.x) is really old and emitting some warnings with php >= 8.1. So, just move to current 4.x version.

Simply change to its new incarnation (fork of the original) and tidy up a few parameters.

Note that the phar still has a good number of limitations:
- Not all checks work (phpcs, behat...).
- Cannot post-install moodle-local_ci (vnu.jar)
- ...

But all those are pre-existing and this new version does not change that.

Note that box requires PHP 8.1 to build the phar, but the phar works perfectly with PHP 7.4 (our min requirement).

Fixes #235

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,10 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v3
 
-      - name: Setup PHP 7.4
+      - name: Setup PHP 8.1
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: 8.1
           ini-values: phar.readonly = false
 
       - name: Build PHAR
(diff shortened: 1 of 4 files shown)


stronk7:
Self merging... towards a release soon...

PR #237, opened 2023-09-08
Enable mustache check to run from phar

stronk7:
We need both the php script and the bundled vnu.jar files to be put in tmp dir and execute/use them from there.

This shouldn't affect normal (non-phar executions). Note that tests cover the non-phar case.


stronk7:
Self-merged.

PR #238, opened 2023-09-11
Enable phpcs and phpcbf checks to run from phar

stronk7:
Note this is very similar to #237 (that **should be merged before this**).

We need the 2 CLI scripts and the CodeSniffer.conf file to be added to the phar, so we can "include" the scripts for execution and copy the conf file temporarily to be together with the phar (that's the way CodeSniffer expects it).

Note this will reduce a little bit code-coverage but in a future PR I want to add some integration tests to cover the execution from phar, hardly can be done from unit tests.

### DIFF
--- a/box.json
+++ b/box.json
@@ -36,6 +36,14 @@
         "Docs",
         "pix"
       ]
+    },
+    {
+      "in": "vendor/squizlabs/php_codesniffer",
+      "name": [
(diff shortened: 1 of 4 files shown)


kabalin (approved):

stronk7:
Thanks, @kabalin , merged!

PR #239, opened 2023-09-11
Fix .env support when running from phar archive

stronk7:
To be **merged after #238**.

With this changes, the moodle-plugin-ci execution
will be able to find a .env file in the same directory that the .phar file containing it.

That way install will be able to create that file and the rest of commands will access to that information via dotenv, avoiding to have to export any env variable manually for phar-based executions.


kabalin (dismissed):

stronk7:
Merged, thanks (I just rebased the branch after your review, @kabalin, and that somehow automatically dismissed your review). I've go ahead.

PR #241, opened 2023-09-15
Add phar integration tests

stronk7:
Over the last days, we have fixed a number of checks to work ok when using the `moodle-plugin-ci.phar` alternative.

Obviously, that's not covered with unit tests, so we need to add some integration tests, similar to the `bin/moodle-plugin-ci` ones, but using the PHAR instead.

Here they are, we test them once for each PHP version supported, that should be enough.

Plus, we make the PHAR available as artifact in the GHA runs so people can download them from there (normally we only rebuild the PHAR for releases).

Ciao :-)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,6 @@ jobs:
         files: ./build/logs/clover.xml
         verbose: true
 
-
   citest:
     name: Integration tests
     needs: selftest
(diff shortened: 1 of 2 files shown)


kabalin (approved):
Thanks @stronk7. Greeting from MoodleMoot!

stronk7:
> Greeting from MoodleMoot!
🧡 🧡 🧡

PR #242, opened 2023-09-24
Update .gitattributes to current files to keep out from packages

stronk7:
We had not updated this file since long ago and there are some new/moved paths to get excluded from distribution when installed as a dependency.


kabalin (approved):
👍

stronk7:
Thanks, @kabalin !

PR #245, opened 2023-09-29
Add docs section about running in local dev environment

kabalin:


stronk7:
Hi Ruslan,

thanks for writing this.

Regarding the PHAR usage, what I do it to put it (a hard-link, I think that there are still some problems with soft links) into all my moodle base directories (note that Moodle's `.gitignore` has `moodle-plugin-ci.phar` defined, so it doesn't disturb.

That, also allows me to just update any of the phar files to current one and bump, all them become updated. In any case this is only a detail.

The important point is that, having the phar in the moodle base directory makes things easier and you can run (from base dir):
```
php moodle-plugin-ci.phar command path/to/plugin
```

Similarly, the `moodle-plugin-ci` binary can be added to the `$PATH` without problems and then it's possible to execute:
```
moodle-plugin-ci command path/to/plugin
```

Of course, for both to work, the need to be executed from base directory. Or, alternatively, specify where the base directory is using `--moodle /path/to/moodle/base`.

So this works:
```
cd enrol (for example)
$ php ../moodle-plugin-ci.phar mustache --moodle=.. ../mod/glossary
$ moodle-plugin-ci mustache --moodle=.. ../mod/glossary
```

Anyway. TL;DR; For your consideration if 1) putting the phar in moodle base dir or 2) adding the binary to the $PATH are good enough to comment about them. Always making clear that, in order for any of them (phar or binary) to work, they must be executed from base dir (dirroot).

Ciao :-)


Not sure if t's worth explain it (that the .phar can be put in base directory) or no. Personally I'm used to that. Maybe because I also follow the very same strategy with `composer.phar`.

stronk7:
Other that that. I like the document. Surely if we start adding more and more details to it (say `.env` files, say other configurations...) we'll need to move to another docs page, but current proposal fits well in the main page, IMO.

stronk7 (dismissed):

kabalin:
Thanks Eloy, I modified the docs slightly to keep `moodle-plugin-ci.phar` within Moodle dir, thanks for the idea. I think usage example with $PATH and hard links may be suitable if we write a separate doc for that one day, let's keep it simple for index page :)

stronk7 (approved):
Approved, thanks!

PR #249, opened 2023-10-11
Add back the selfupdate command (interim mode)

stronk7:
The self update command dependencies were removed in the past but the command was left around. As a result of that any attempt of running:

`php moodle-plugin-ci.phar selfupdate`

Is failing with PHP error (because the underlying humbug/phar-updater library was not available any more.

So, with this commit, we are re-introducing it, but with the following observations:

1. We have moved from humbug/phar-updater (archived) to laravel-zero/phar-updater that is maintained.
2. We are stuck to version 1.*, because of we still supporting PHP 7.4
3. I've run it with PHP 8.2 and it doesn't break badly, just some PHP deprecated notices are displayed. In any case, far better than the previous status, not working at all.
4. Having it back to code will allow us to further progress:
   - Getting it tested with some new integration test.
   - Have time to look if there is any other fork with relaxed requirements (or make our own one).
   - Try other selfupdate alternatives. Certainly the one we are currently using is the most advanced ones, although for basic usage (no rollback, no keys verification...) there are a few, with `consolidation/self-update` surely being the more active (again knowing that it's way simpler).

And that's the story of this PR, basically just keep the selfupdate command working (knowing that emits a few PHP notices with PHP 8.x) but allowing us to test it and try other alternatives in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,8 @@
     "psr/log": "^1.1.4",
     "nikic/php-parser": "^4.14",
     "marcj/topsort": "^2.0.0",
-    "phpcompatibility/php-compatibility": "dev-develop#0a17f9ed"
+    "phpcompatibility/php-compatibility": "dev-develop#0a17f9ed",
+    "laravel-zero/phar-updater": "^1.0.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^9.6",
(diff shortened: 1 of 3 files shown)


stronk7:
Note: If you want to check how it works, you can download the generated `moodle-plugin-ci.phar` file from the GHA run and then try with it. Obviously it will tell you that you are already updated, only when we make a new release it will start to detect and self updating.

Alternatively, also can build it via `make clean && make build` and the new phar version will be generated under the `build`directory.

PR #250, opened 2023-10-11
Add tests to cover what is testable in SelfUpdate command

stronk7:
Note that only helper methods can be tested for that command, not the execution of the update itself (it has to happen within a PHAR). And it doesn't make much sense to start mocking lots of stuff for that.

Some integration tests @ CIS will be in charge of
covering the real execution instead. Coming soon.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -261,3 +261,45 @@ jobs:
         php build/moodle-plugin-ci.phar behat --profile default
         php build/moodle-plugin-ci.phar behat --profile chrome
         php build/moodle-plugin-ci.phar behat --profile firefox --tags="@local_ci&&~@app"
+
+  selfupdatetest:
+    name: SelfUpdate tests (PHAR)
+    runs-on: ubuntu-22.04
+    env:
+      lowest_release: '4.1.8'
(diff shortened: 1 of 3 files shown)


kabalin (dismissed):

stronk7:
Thanks @kabalin,

I'm going to keep this on hold until we have a couple of releases with the self-update command working. Only then we'll be able to add some integration (GHA) test to verify that it's working as expected.

Ciao :-)

stronk7:
Ok, now that we have > 2 releases supporting the `selfupdate` command, we can add some simple integration tests for it.

I've added a second commit that just:
- downloads the 4.1.8 (first version with the command working) `moodle-plugin-ci.phar` asset.
- runs `selfupdate` to become updated to actual.
- runs `selfupdate --rollback` to go back to the original 4.1.8.

They seem to be passing... so I'm going to make this PR ready for review, it includes both the original unit tests and the new integration tests.

Ciao :-)

stronk7:
I'm sef-merging this upstream now, just to have the `selfupdate` command covered and detect any potential regression, specifically when bumping various components versions.

Ciao .-)

PR #251, opened 2023-10-11
ci examples: use !cancelled() instead of always()

### LINKED ISSUE #248: GitHub Action: Avoid "if: always()"
From https://docs.github.com/en/actions/learn-github-actions/expressions#always

> If you want to run a job or step regardless of its success or failure, use the recommended alternative: `if: ${{ !cancelled() }}`

In https://github.com/ncstate-delta/moodle-mod_zoom/pull/531, I proposed change to our plugin's GitHub Action. It occurred to me that some of these changes might be beneficial to include in the example file `gha.dist.yml`. Here are some of the benefits it may offer:

1. Only run moodle-plugin-ci steps when the moodle-plugin-ci installs successfully
2. Use `!cancelled()` instead of `always()` as recommended by GitHub
3. If the job has been cancelled, mark the job as failed so that PR status checks cannot be bypassed

If you would prefer, I would be willing to open a PR. Mainly, I want to make sure I'm only suggesting desired/acceptable changes.

jrchamp:
As described in #248, GitHub discourages the use of always() when it can be avoided.
Fixes #248

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,13 +9,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Changed
+- ACTION SUGGESTED: If you are using GitHub Actions, it's recomended to use `!cancelled()` instead of `always()` for moodle-plugin-ci tests. Adding a final step that always returns failure when the workflow is cancelled will ensure that cancelled workflows are not marked as successful. For a working example, please reference the updated `gha.dist.yml` file.
+
 ## [4.1.8] - 2023-10-20
 ### Changed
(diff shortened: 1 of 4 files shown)


stronk7:
Hi @jrchamp ,

many thanks for the patch! For me it looks perfectly ok to start suggesting to use `!cancelled()`.

Only suggestion I have is... maybe you can add a note about the change @ [docs/CHANGELOG.md](https://github.com/moodlehq/moodle-plugin-ci/blob/master/docs/CHANGELOG.md), under a new `Unreleased -> Changed` section, with an `ACTION SUGGESTED: xxxx` item explaining it ?

Ciao :-)

jrchamp:
Thanks @stronk7! I've added an item to the docs/CHANGELOG.md file.

stronk7:
Perfect, @jrchamp, here we go!

stronk7 (approved):

PR #255, opened 2023-11-04
Add the libonig5 to the Travis packages to install

stronk7:
For some strange reason, travis has started to fail with PHP 8.2 with error:

php: error while loading shared libraries: libonig.so.5

It was not happening few days ago. Searching out there it seems that the fix is as simple as apt-installing that library.

It's incredible that they don't have that implemented in their environments and things like this need to be installed manually.

Any way, fixed in our own Travis test and also in the template and docs.

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -8,6 +8,7 @@ addons:
     packages:
       - postgresql-13
       - postgresql-client-13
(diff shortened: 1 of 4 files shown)


paulholden (approved):
LGTM

PR #259, opened 2023-12-06
Implement --exclude option in phpcs command

NoelDeMartin:
I've been trying to exclude some rules from a plugin using a `.phpcs.xml` file and I didn't manage to do it. But it was very easy using [the exclude option](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#limiting-results-to-specific-sniffs). I think this would be a nice addition.

### DIFF
--- a/src/Command/CodeCheckerCommand.php
+++ b/src/Command/CodeCheckerCommand.php
@@ -39,6 +39,7 @@ class CodeCheckerCommand extends AbstractPluginCommand
             ->setAliases(['codechecker'])
             ->setDescription('Run Moodle CodeSniffer standard on a plugin')
             ->addOption('standard', 's', InputOption::VALUE_REQUIRED, 'The name or path of the coding standard to use', 'moodle')
+            ->addOption('exclude', 'x', InputOption::VALUE_REQUIRED, 'Comma separated list of sniff codes to exclude from checking', '')
             ->addOption('max-warnings', null, InputOption::VALUE_REQUIRED,
                 'Number of warnings to trigger nonzero exit code - default: -1', -1)
             ->addOption('test-version', null, InputOption::VALUE_REQUIRED,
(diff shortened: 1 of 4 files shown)


stronk7:
Sold!

src/Command/CodeCheckerCommand.php
```diff
@@ -39,6 +39,7 @@ protected function configure(): void
             ->setAliases(['codechecker'])
             ->setDescription('Run Moodle CodeSniffer standard on a plugin')
             ->addOption('standard', 's', InputOption::VALUE_REQUIRED, 'The name or path of the coding standard to use', 'moodle')
+            ->addOption('exclude', null, InputOption::VALUE_REQUIRED, 'Comma separated list of sniff codes to exclude from checking', '')
```
stronk7, line 42: This is the only detail I'd argue something, maybe to keep it aligned with the `--standard` option that has the short `-s` equivalent... maybe I'd add `-e` for `--exclude`.

stronk7 (commented):
Up to you the little details commented, just tell me what are you going to do (ignore or add it) and I'll approve this.

Thanks!

src/Command/CodeCheckerCommand.php
NoelDeMartin: I thought about it, but there is currently an `-e` flag for the standalone `phpcs` command (the one you run from php, not from moodle-plugin-ci) that means something else: "[Explain a standard by showing the sniffs it includes](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage#getting-help-from-the-command-line)".

So I thought it could be misleading to have it here (an even potentially create issues in the future if we decide to implement that option as well).

But if you still think we should add it, I'm fine with it. Wouldn't be too much of an issue.

NoelDeMartin: Okay, done!

stronk7:
Sold, thanks!

PR #267, opened 2024-01-17
Prevent plugin runs to use the moodle.Commenting.TodoComment sniff

### LINKED ISSUE #266: Ensure that we only use the TodoComment Sniff for core runs, not for plugin ones
With https://github.com/moodlehq/moodle-cs/pull/91 we are introducing a new Sniff that looks for missing MDL-xxxx issue codes in all the TODO and @todo comments (as stated by the coding style).

This Sniff is only for core, hence we have to prevent it to be executed for plugins within moodle-plugin-ci. By disabling it or by setting `commentRequiredRegex` to an empty value.

Ciao :-)

stronk7:
A new `todo-comment-regex` option has been added to the phpcs command.

It allows to specify the regex that will be used with inspecting
todo (TODO and @todo) comments.

By default, an empty string is used for the option and that makes
the Sniff to stop checking. Whoever wants to check for anything
(tracker issue, github issue, arbitrary url, ...) can us the new
option to configure it.

Fixes #266

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Added
+- Added support for the `--todo-comment-regex` option to the `phpcs` command. When specified, all the todo comments (`TODO` and `@todo`) are inspected to ensure that they contain some text matching the regular expression (a tracker issue key: `MDL-[0-9]+`, a link to GitHub: `github.com/moodle/moodle/pull/[0-9]+`, ... or any other valid alternative).
+
 ### Changed
 - Updated project dependencies to current [moodle-cs](https://github.com/moodlehq/moodle-cs) and [moodle-local_ci](https://github.com/moodlehq/moodle-local_ci) versions.
 
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -1572,7 +1572,7 @@ Run Moodle CodeSniffer standard on a plugin
 
 ### Usage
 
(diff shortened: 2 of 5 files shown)


stronk7:
Commit amended, now the new `todo-comment-regex`takes the baton to decide how to proceed:

- By default it uses empty string value that, at all effects, disables the todo Sniff completely.
- By setting it to anything else, the todo Sniff will check that all the todo comments (TODO and @todo), match the regex.

Ciao :-)

kabalin (approved):
Thanks Eloy, LGTM.

kabalin:
Changelog is record is required (feel free to add later).

stronk7:
Thanks @kabalin,

I'll add the CHANGELOG details, yep (tomorrow).

Also I've converted this to draft because we need the new moodle-cs release to happen in order to get the new unit tests that I've added passing (they require a working moodle-cs with the new config option).

Once we bump to new moodle-cs, I'll complete this.

Ciao :-)

stronk7:
CHANGELOG updated with text:

### Added
- Added support for the `--todo-comment-regex` option to the `phpcs` command. When specified, all the todo comments (`TODO` and `@todo`) are inspected to ensure that they contain some text matching the regular expression (a tracker issue key: `MDL-[0-9]+`, a link to GitHub: `github.com/moodle/moodle/pull/[0-9]+`, ... or any other valid alternative).

Still holding on https://github.com/moodlehq/moodle-cs/pull/91 and `moodle-cs` release, to progress this. Not in a hurry.

stronk7:
About to go merging this (once tests end), as far as it was approved with all the changes but the `CHANGELOG` ones that came later.

Immediately after getting this landed, will make a release, so it can be tested out there (default and custom regular expressions).

Ciao :-)

Issue #15, opened 2020-07-17
Add the ability to configure which firefox / chrome versions (selenium images) we want to use


stronk7:
Recently we have started to run behat tests using dockerized browser + selenium. Right now, and until we get the webdriver drives updated to newer W3C ones in behat... we are using super-old Firefox 2.53.1 and, for Chrome, we are running on the edge (latest).

Whenever we change the drivers to the new ones, or if there are breaking changes in chrome (has happened a number of times), of if somebody wants to run the tests against older versions, it should be possible to specify/pass the exact version  to use (or maybe whole image if a custom one, say `orangehq/super-selenium:99.99`, not sure, just sharing the thought).

This was created while reviewing #13.

PR #268, opened 2024-01-31
Allow selenium image configuration

NoelDeMartin:
I've decided to remove the private properties because they were only used once, and now everything is encapsulated in the `getSeleniumImage` method.

Other than that, there isn't much more to the PR. It allows to configure the selenium image by setting `MOODLE_BEHAT_SELENIUM_IMAGE`, or specify it for each profile with `MOODLE_BEHAT_SELENIUM_CHROME_IMAGE` and `MOODLE_BEHAT_SELENIUM_FIREFOX_IMAGE`.

I also pondered whether to allow indicating only the browser version. For example, setting `MOODLE_BEHAT_CHROME_VERSION` or accepting `chrome:117` for the profile. But I wasn't convinced by that approach, so I ended up using this one that also allows customizing the image completely (for example, in case [you'd need to use seleniarm](https://github.com/moodlehq/moodle-docker/pull/250) or something else).

### DIFF
--- a/src/Command/BehatCommand.php
+++ b/src/Command/BehatCommand.php
@@ -26,25 +26,6 @@ class BehatCommand extends AbstractMoodleCommand
     use ExecuteTrait;
 
     /**
-     * Selenium legacy Firefox image.
-     */
-    private string $seleniumLegacyFirefoxImage = 'selenium/standalone-firefox:2.53.1';
-
-    /**
-     * Selenium standalone Firefox image.
-     *
-     * @todo: Make this configurable.
-     */
-    private string $seleniumFirefoxImage = 'selenium/standalone-firefox:3';
-
-    /**
-     * Selenium standalone Chrome image.
-     *
-     * @todo: Make this configurable.
-     */
-    private string $seleniumChromeImage = 'selenium/standalone-chrome:3';
-
-    /**
      * Wait this many microseconds for Selenium server to start/stop.
      *
      * @var int<0,max>
@@ -65,6 +46,7 @@ class BehatCommand extends AbstractMoodleCommand
             ->addOption('name', null, InputOption::VALUE_REQUIRED, 'Behat name option to use', '')
             ->addOption('start-servers', null, InputOption::VALUE_NONE, 'Start Selenium and PHP servers')
             ->addOption('auto-rerun', null, InputOption::VALUE_REQUIRED, 'Number of times to rerun failures', 2)
+            ->addOption('selenium', null, InputOption::VALUE_REQUIRED, 'Selenium Docker image')
             ->addOption('dump', null, InputOption::VALUE_NONE, 'Print contents of Behat failure HTML files')
             ->setDescription('Run Behat on a plugin');
     }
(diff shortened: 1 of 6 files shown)


stronk7:
I like to be able to specify the images... just guessing when we should follow the `env` route here or just add one option (`--selenium` to the command (that, of course, will take precedence over the profile one).

Right now it's really a mix of env and options, grrr. Need to think a little bit more about it...

stronk7:
OT: @NoelDeMartin, I've force-pushed (without changes) your candidate branch to get the tests executed, it seems that some of them were stuck.

stronk7:
> It seems like some tests are failing in Travis but I cannot see the details.

Yeah, Travis is being a nightmare since 2 weeks ago. Sometimes it installs PostgreSQL on port 5432, others in 5432... I've been fighting with that in #270 , just to, at the end, leave the port like it's now (5432).

Will look to this soon... thanks!

stronk7:
Side note, surely this closes #15

stronk7:
Thanks @NoelDeMartin,

it's 98% perfect for me! The remaining 2% is:

1. Do we really need the 3 env variables? Why isn't the "general" (`MOODLE_BEHAT_SELENIUM_IMAGE`) one enough for everything? Whoever wants to set a custom selenium image just does it using it (or the new option) and done. I've tried to imagine use cases and I'm not sure to get the point about the browser specific ones.
2. No matter the previous point outcome, I think we need to document a little bit more the env variables. Maybe in the changelog, or maybe in the `docs/Help.md` page (there is a Behat section there).

Thoughts?

NoelDeMartin:
Thanks for the review @stronk7, I have applied your suggestions:

1. I just thought it'd made sense to override them individually given that we have 2 different profiles and the same image couldn't work for both. But I don't mind removing them. Most people is likely to use only one profile, and if they don't they can now pass a different `--selenium` flag.
2. There are other env variables such as `MOODLE_START_BEHAT_SERVERS` that are not documented, so I wasn't sure if that was necessary (specially now that we have the documented `--selenium` flag). But I added it in the CHANGELOG.

stronk7:
> There are other env variables such as MOODLE_START_BEHAT_SERVERS that are not documented,

Yeah, I was looking for it in the Docs and only the "quick start" section of https://moodlehq.github.io/moodle-plugin-ci/Help.html was showing some options.

I've created #271 about that, surely we'll need to make a few new pages (similar to the app one) and document many of those environmental variables. In any case that falls out of this PR scope.

Ciao :-)

stronk7:
I will merge this as soon as we are green (tests still running). Thanks for your work and patience!

src/Command/BehatCommand.php
```diff
@@ -65,6 +46,7 @@ protected function configure(): void
             ->addOption('name', null, InputOption::VALUE_REQUIRED, 'Behat name option to use', '')
             ->addOption('start-servers', null, InputOption::VALUE_NONE, 'Start Selenium and PHP servers')
             ->addOption('auto-rerun', null, InputOption::VALUE_REQUIRED, 'Number of times to rerun failures', 2)
+            ->addOption('selenium', null, InputOption::VALUE_NONE, 'Selenium Docker image')
```
stronk7, line 49: Oh, sorry for not detecting this earlier. I don't think that `InputOption::VALUE_NONE` is correct here. If the option is used, a value (valid selenium image) is required.

stronk7: I'm not sure why the tests are passing, somehow thy are bogus and not validating the `VALUE_NONE` restriction that should avoid that option to be passed with any value.

stronk7 (changes requested):

src/Command/BehatCommand.php
NoelDeMartin: Thanks for spotting that, it seems like I hadn't completely understood how those constants worked. I fixed it now.

Not sure about how to improve the tests, though. But I guess it's not related to this feature so if you think it's important to look into that I'd suggest opening a separate issue.

stronk7: Yeah, will check why the tester is not catching that `VALUE_NONE` "violation". It should. But it's apart from this PR, 100% agree.

stronk7:
Sold!

PR #269, opened 2024-02-07
Configure browser capabilities

NoelDeMartin:
I have tried using the [add-config](https://moodlehq.github.io/moodle-plugin-ci/AddExtraConfig.html) command to achieve this instead, but it seems like the Behat configuration is created in the `install` command so changes done afterwards are not applied. And if I try to use the `add-config` command before installing, I get an error.

So the only solution I've found is to modify `config.php.txt` before running the `install` command, but that's a bit fickle so it would be nice to have an idiomatic way to achieve it.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 
 ### Added
 - `--selenium` option or `MOODLE_BEHAT_SELENIUM_IMAGE` env variable to `behat` to specify Selenium Docker image.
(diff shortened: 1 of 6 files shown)


stronk7:
Maybe we could add some test applying for capabilities and check the results in config file? If it's complex to do, please create an issue to do it apart from this. You decide, 100% (I've not looked how achievable it is).

Ciao :-)

NoelDeMartin:
Thanks @stronk7, I've added a couple of tests duplicating the config file fixture, I think that's the easiest way.

stronk7 (approved):
LGTM, thanks!

PR #270, opened 2024-02-07
Show a warning annotation when "master" branch usage is detected

stronk7:
That way, both the web UI and the email from GitHub will show the recommendation about to move to the new "main" branch.
    
Also, unrelated, exclude some unreachable code from code coverage analysis.

Here there is an example to show how the annotation looks like (at the bottom):

https://github.com/stronk7/moodle-plugin-ci/actions/runs/7812729398

Note that an unrelated 2nd commit has been added about to try to keep Travis working, because it has started to fail recently (maybe they changed the PostgreSQL ports again, grrr).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,14 +9,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
-
 ### Changed
 - Updated all uses of `actions/checkout` from `v3` (using node 16) to `v4` (using node 20), because [actions using node 16 are deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) and will stop working in the future.
-* ACTION SUGGESTED: In order to avoid the node 16 deprecation warnings, update your workflows to use `actions/checkout@v4`.
+- ACTION SUGGESTED: In order to avoid the node 16 deprecation warnings, update your workflows to use `actions/checkout@v4`.
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


stronk7:
Side note, I'm not sure what's going with Travis since 1-2 weeks ago, maybe they are upgrading their environments because all our PostgreSQL related tests are failing with different errors (not package available, not running @ port xxxx, ...).

I'll take a look to that, but apart from this issue, just sharing why it's failing.

Ciao :-)

stronk7:
Other than that, I think this is ready for review as far as it seems to be working ok. Feel free to suggest a better wording or whatever (I'm not good there!).

Ciao :-)

stronk7:
Ok,

confirmed, Travis has changed PostgreSQL default port from the exceptional 5432 that it had since last year to the usual 5433 that they have for all versions since v10.

This run (that uses the 2nd commit added here) is now running ok:

https://app.travis-ci.com/github/moodlehq/moodle-plugin-ci/builds/268805831

So, if we get everything green, we can go with this. 2 issues fixed in 1 PR.

Ciao :-)

stronk7:
Grrr, it's flip.flopping (the PostgreSQL 13 port). Last run failed with the change back to 5433. And now it's showing 5432 again:

```
$ ls -al /var/run/postgres*
total 8
drwxrwsr-x  3 postgres postgres  120 Feb  8 15:42 .
drwxr-xr-x 34 root     root     1100 Feb  8 15:42 ..
drwxr-s---  2 postgres postgres   80 Feb  8 15:42 13-main.pg_stat_tmp
-rw-r--r--  1 postgres postgres    5 Feb  8 15:42 13-main.pid
srwxrwxrwx  1 postgres postgres    0 Feb  8 15:42 .s.PGSQL.5432
-rw-------  1 postgres postgres   69 Feb  8 15:42 .s.PGSQL.5432.lock
```
(yesterday it was showing 5433, that's the reason I added the 2nd commit)

I'm going to switch back to 5432, removing the second commit in this PR and doing a few more tests... grrr.

Ciao :-)

kabalin (approved):
LGTM 👍

stronk7:
Thanks Mr. @kabalin !

PR #273, opened 2024-02-18
Validation of dataformat - custom plugin name lang string

marinaglancy:
`moodle-plugin-ci validate` fails on dataformat plugins because they use language string 'dataformat' as a plugin name.


kabalin:
> I don't know if I should change anything in the file headers - they mention "Copyright (C) Blackboard"

I think put yourself there, [Eloy does](https://github.com/moodlehq/moodle-plugin-ci/blob/main/tests/Command/SelfUpdateCommandTest.php#L9).

marinaglancy:
I changed the copyrights

src/PluginValidate/Requirements/DataformatRequirements.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace MoodlePluginCI\PluginValidate\Requirements;
+
+use MoodlePluginCI\PluginValidate\Finder\FileTokens;
+
+/**
+ * Dataformat plugin requirements.
+ */
+class DataformatRequirements extends GenericRequirements
```
kabalin, line 20: Should we be more strict and check `classes/writer.php` existence too?

stronk7:
pong!

src/PluginValidate/Requirements/DataformatRequirements.php
marinaglancy: I think if we add proper validation to the plugins, we should start with the most used ones. This PR is only about removing the false positive error

kabalin (approved):

PR #274, opened 2024-02-21
Update App config defaults

NoelDeMartin:
I'm opening this PR as a draft because until v4.4.0 of the app is released, these defaults don't work with the `moodlehq/moodleapp:latest-test` image, which is the default (they only work with `moodlehq/moodleapp:next-test`). But we can get the review started so that once 4.4.0 is released, we can merge this.

`moodle-docker` has a similar issue, but given that we explicitly support different versions, I've implemented an interim solution in https://github.com/moodlehq/moodle-docker/pull/281.

But for `moodle-plugin-ci`, I think it's fine just to support one version out of the box. And anyone can configure old versions if they want using other env variables (the documentation has also been updated).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,14 +77,19 @@ jobs:
         include:
           - php: '8.3'
             moodle-branch: 'main'
+            moodle-app: 'true'
           - php: '8.3'
             moodle-branch: 'MOODLE_404_STABLE'
+            moodle-app: 'true'
           - php: '8.2'
             moodle-branch: 'MOODLE_403_STABLE'
+            moodle-app: 'true'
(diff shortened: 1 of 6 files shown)


stronk7 (approved):
Thanks @NoelDeMartin ,

really, there is something with the relations between `MOODLE_APP_PORT`, `MOODLE_APP_PROTOCOL` and `MOODLE_BEHAT_IONIC_WWWROOT` that I'm unable to understand completely, but, hey, that doesn't matter much for the purposes of the PR.

So, as far as it's all passing, I'm ok. I imagine that 99% of people really only needs to enable `MOODLE_APP` and that's all.

Ciao :-)

PR #278, opened 2024-03-07
Limit stylelint tasks to plugin directory

### LINKED ISSUE #256: Add support for limiting scope of grunt
Use case:
When additional plugins needs to be installed by moodle-plugin-ci because the plugin I want to run the pipeline for depends on some other plugins, I want to run grunt and all other jobs exclusively for my plugin, NOT for the dependency plugins.

Currently however, grunt for example checks all installed plugins which means also the installed plugins just needed as dependency.

There need to be a config option (or even change the default behavior) which limits the scope to the plugin itself, not the dependency plugins.

PhMemmel:
Closes #256

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- Fixed a problem with the `grunt` command running the `stylelint` tasks against the whole Moodle directory (including both core and other optional plugins installed). Now only the plugin being checked is effectively analysed.
+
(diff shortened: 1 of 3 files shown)


stronk7:
Hi @PhMemmel ,

I've put this a little bit on hold until we discuss @ #256 . Thanks for your patience!

stronk7:
Hi @PhMemmel ,

I've taken the liberty of amending your commit, fixing some styling details, one test that was failing and adding the information to the changelog file.

Hope that's ok for you. If you're happy, I'll be merging this once we get all checks passing.

Ciao :-)

PhMemmel:
Feel free to add whatever is necessary! Thank you very much for taking over!

stronk7 (approved):

stronk7:
Merged, let's see if we can roll this soon... thanks @PhMemmel !

Issue #262, opened 2023-12-12
Remove Copy/Paste Detector integration


stopfstedt:
during `composer install`, i'm seeing this warning:

```
Package sebastian/phpcpd is abandoned, you should avoid using it. No replacement was suggested.
```

and true enough, the code repo has been flagged as abandoned. https://github.com/sebastianbergmann/phpcpd

i'm proposing the complete removal of this integration.

stronk7:
Hi @stopfstedt ,

yeah, we were aware about the removal. Also, we have been monitoring how forks (of phpcpd) are evolving, with a couple of them that "may" be interesting to switch to.

Thanks for creating this and #263, we'll analyse this soon, because one of the things we need to carefully consider is how the (potential) removal is going to affect to all the plugins out there that are running the `phpcpd` command, maybe requiring it to stay, but warning about the deprecation of the command, instead of a sudden deletion.

Let's see how everything evolves... I'll keep both this and the PR open as a reminder... hope we can come with a decision soon (switch, deprecate, remove...).

Ciao :-)

stopfstedt:
@stronk7 thanks for the review.
i agree with your concerns - sudden removal might be a bit jarring.

i started with a new PR that deprecates this feature as an intermediary solution.

stronk7:
Now that the deprecation has landed as part of #264 , I'll keep this open for whenever we make the next major version (5.0.0). Then, the `phphpd` command will be completely removed.

Ciao :-)

jrchamp:
@stronk7 Now that `phpcpd` is deprecated, it seems like it should be removed from the sample workflow files. I understand that complete removal should happen in a major version (because it breaks backward compatibility), but the sample workflow files are more like "suggested best practices", which should probably be updated to stop using deprecated commands. With that said, whatever you decide, I'm fine either way.

stronk7:
> it seems like it should be removed from the sample workflow files.

Agreed, I've created #290 to remove it from dist and "explain" docs. Good idea!

Ciao :-)

stronk7:
Ok, I' think that this can be considered done.

The complete removal has happened as part of the https://github.com/moodlehq/moodle-plugin-ci/milestone/1 , and now it's part of the https://github.com/moodlehq/moodle-plugin-ci/tree/wip_v5 branch.

Ciao :-)

PR #290, opened 2024-03-16
Remove phpcpd (deprecated) from dist and explain docs

stronk7:
Better stop having the deprecated phpcpd command available in the dist and explain docs, so newcomers won't enable it by default. It's going to be removed in v5.

Part of #262, credit goes to @jrchamp

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -51,7 +51,6 @@ install:
 
 script:
   - moodle-plugin-ci phplint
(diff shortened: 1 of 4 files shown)


kabalin (approved):

kabalin:
LGTM

stronk7:
Wow, that was quick! Thanks @kabalin !

stronk7:
Also merging it quickly! Heh!

PR #291, opened 2024-03-16
Add step for Behat faildump storing as GHA artifact.

kabalin:
Add GHA step to store Behat faildump as workflow artifact, so it can be [inspected](https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts).

image

This step will be run only if Behat step failed (see `if` statement). It also configured not to produce warning if files are not found (the cause of Behat error might be not the actual test scenario). Name is derived from the job matrix, it needs to be unique per job when matrix is used (artifact upload will fail on name conflict if more than 1 job will fail on Behat)

Demo (simplified for speed):
* Failed job: https://github.com/kabalin/moodle-format_flexsections/actions/runs/8307608604
* Succeeded job (no artifacts): https://github.com/kabalin/moodle-format_flexsections/actions/runs/8307723810

More info on upload artifact: https://github.com/actions/upload-artifact

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+
+- Added GHA step to store Behat faildump as workflow artefact, so it can be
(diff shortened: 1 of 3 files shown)


stronk7 (approved):
LGTM! 😛

kabalin:
BTW, kudos also go to @marinaglancy who suggested the similar idea for our internal GitLab jobs 🎉

PR #296, opened 2024-03-31
Be stricter and don't allow warnings on phpcs tests

stronk7:
Just ensure that we fail on self integration tests (phpcs) warnings, so we can monitor the impact of every new moodle-cs release better.


stronk7:
I'm going to self-merge this. It's just self tests becoming stricter so we can add some more cases to our "fixture" local_ci plugin and check better how moodle-cs changes behave and impact.

Issue #289, opened 2024-03-15
Remove phpdbg support


stronk7:
This is blocked by https://github.com/moodlehq/moodle-plugin-ci/issues/213 that needs to be implemented first in v4 (deprecation).

But, for sure `phpdbg` doesn't work with PHPUnit 10 and up anymore, so we must drop support in v5 yes or yes.

Ciao :-)

PR #298, opened 2024-04-06
Deprecate the phpdbg option for code coverage

### LINKED ISSUE #213: Deprecate phpdbg support
Since PHP 8.0 `phpdbg` is causing more and more problems, like random crashes (#204). And things are worse in PHP 8.1... (in fact I think I've read somewhere that it was going to be removed, but I've failed to find the link).

In any case, the fact is that it is suboptimal when used for calculating code-coverage (the use we have of it). And, luckily, there are 2 good alternatives: `xdebug 3` (any version) and `pcov` (Moodle 3.10 and up), both supported by us since June 2022. Both are nicely documented, and easy to setup, see #212).

So this issue is about to agree to remove `phpdbg` support and do it.

Ciao :-)

stronk7:
This includes both the current default (when neither pcov or xdebug are available) and the explicit use of --coverage-phpdbg

Since PHP 8.0, phpdbg is causing all sort of problems and is not recommended. So it's time to deprecate it in order to proceed to remove it completely for moodle-plugin-ci 5.0.0.

Removal ref.: https://github.com/moodlehq/moodle-plugin-ci/issues/289

Fixes #213

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
+### Deprecated
+- The use of `phpdbg` to calculate PHPUnit's code-coverage has been deprecated in this `moodle-plugin-ci` release (4.5.0) and will be removed in 5.0.0. This includes both the implicit (default) option when no alternative (`pcov` or `xdebug`) is available and the explicit `--coverage-phpdbg` option.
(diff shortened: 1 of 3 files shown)


stronk7:
Just for testing purposes, this is how the deprecation warning notification is displayed whenever `phpdbg` is effectively used (note that only old php < 8.0 jobs can be using it, as commented, any 8.x job will fail badly).

https://github.com/stronk7/moodle-plugin-ci/actions/runs/8582321848

kabalin (approved):
LGTM 🚀 Also thanks for adding changelog record for dataformat (overlooked, sorry).

stronk7:
Ta, K!

Issue #297, opened 2024-04-01
Grunt job is not ignoring hidden directories


ewallah:
The grunt job is not ignoring hidden directories. The job copies the complete directory of a plugin to a temporary space, does an excellent job and mirrors is back. If you have for example a .git folder with special directory permissions, the restore fails with a permission denied error.

```
Failed to copy "/tmp/moodle-plugin-ci-grunt-backup-1711963880/.git/objects/57/..." to ".../.git/objects/57/..." because target file could not be opened for writing: fopen(../.git/objects/57/...): Failed to open stream: Permission denied
```
Therefore I changed the restorePlugin method
 ```
        $flags = \FilesystemIterator::SKIP_DOTS;
        $iterator = new \RecursiveDirectoryIterator($this->backupDir, $flags);
        (new Filesystem())->mirror($this->backupDir, $this->plugin->directory, $iterator, ['delete' => true, 'override' => true]);
```

malteschmitz:
I do have the exact same issue of `Failed to copy ...: Failed to open stream: Permission denied` because `restorePlugin()` tries to restore something deep inside the `.git` structure.

Unfortunately your fix is not solving the problem but breaking the entire backup mechanism. It will randomly delete files, because Symfony's `mirror` is broken if you combine the `iterator` argument with the `'delete' => true` option: https://github.com/symfony/symfony/issues/14068

If you use `mirror` with the `iterator` argument set to null (as the original `restorePlugin` implementation does), then it internally creates two different iterators for the deletion and the copying. However, both of these iterators already include the `SKIP_DOTS` flag: https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Filesystem/Filesystem.php#L522

Finally, as pointed out by [stronk7 in this comment](https://github.com/moodlehq/moodle-plugin-ci/pull/301#issuecomment-2162661607): The SKIP_DOTS flag is only about skipping `.` and `..` and not other files starting with a dot.

PR #301, opened 2024-05-03
Issue #297

ewallah:
Grunt job is not ignoring hidden directories #297

### DIFF
--- a/src/Command/GruntCommand.php
+++ b/src/Command/GruntCommand.php
@@ -112,7 +112,9 @@ class GruntCommand extends AbstractMoodleCommand
      */
     public function restorePlugin(): void
     {
-        (new Filesystem())->mirror($this->backupDir, $this->plugin->directory, null, ['delete' => true, 'override' => true]);
+       $flags = \FilesystemIterator::SKIP_DOTS;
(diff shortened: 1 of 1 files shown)


stronk7:
Hi @ewallah ,

it seems that there are a few coding-style little changes to apply, plus unit tests are failing with your patch.

Is it ok if I amend your branch to make it CI compliant? Other than that, I think that the approach is correct and we can skip dot files without problem (at very least for grunt purposes).

stronk7:
Oh, @ewallah,

I just was adding here a test and confirmed that {{FilesystemIterator::SKIP_DOTS}} does not skip all dot files (.something.txt, .git, ...) but, exclusively, the "." and ".." directories. Hidden files and folders continue being returned by the iterator.

So, this doesn't seem to be enough to achieve what was desired (skip hidden dirs).

ewallah:
Weird, since I added the SKIP_DOTS, I do not see any permission errors any more.

Feel free to add/change whatever is necessary!

malteschmitz:
See my comment on https://github.com/moodlehq/moodle-plugin-ci/issues/297: This fix does not solve the problem but instead breaks the entire backup mechanism.

PR #302, opened 2024-05-03
Add support for MSSQL

### LINKED ISSUE #92: Request: MSSQL support
[Moodle supports more databases](https://docs.moodle.org/dev/Moodle_3.10_release_notes#Database_requirements) than what moodle-plugin-ci currently offers. In https://github.com/frankkoch/moodle-mod_studentquiz/issues/314 we got the first issue for MSSQL.

We would like to support MSSQL, but to confidently do that, we need to ensure that the testing toolchain can test against it as well.

Is moodlehq willing to add support for MSSQL?

If we are lucky, we only need to figure out how travis or github actions files need to be written.

justusdieckmann:
Hey everyone,

this PR adds support for MSSQL (Fixes #92)

I tested it with our plugin tool_lifecycle [here](https://github.com/justusdieckmann/moodle-tool_lifecycle/commits/test/mssql/?since=2024-05-03&until=2024-05-03). The errors in the first commit are correct, since we used some MSSQL-incompatible syntax in the plugin. Those were addressed in the next two commits.

Justus

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -917,7 +917,7 @@ Path to Moodle plugin
 
 #### `--db-type`
 
-Database type, mysqli, pgsql or mariadb
+Database type, mysqli, pgsql, mariadb or sqrsrv
 
 * Accept value: yes
 * Is value required: yes
--- a/gha.dist.yml
+++ b/gha.dist.yml
@@ -27,12 +27,21 @@ jobs:
           - 3306:3306
         options: --health-cmd="mysqladmin ping" --health-interval 10s --health-timeout 5s --health-retries 3
 
+      mssql:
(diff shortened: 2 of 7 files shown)


justusdieckmann:
Sadly, there is no good way around the need for a password for the database with the official docker image, so I passed that to `moodle-plugin-ci install` only if `matrix.database == 'sqlsrv'`.

I have not looked into the .travis.dist.yml yet, since I am much more familiar with github actions.

Also, in the `gha.dist.yml`, I have completely replaced the `${{ matrix.extensions }}` with `sqlsrv`, since `extensions` wasn't set anywhere.

I would be glad for any comments and opinions! :)

PR #304, opened 2024-06-03
Exit on error when something was wrong executing the command

stronk7:
When something abnormal happens (wrong configuration, missing stuff, connectivity problems, etc., the phpcs execution can end with error.

But, if max-warnings is enabled (>=0), then the count of errors and warnings (that is always 0 on error) gets precedence, hiding the real exit status. Hence it passes, when really it was not executed.

Now we detect those abnormal situations and exit with error when something in the execution goes wrong.

Covered with tests, checking all the possible max-warnings combinations.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,6 +15,7 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 
 ### Fixed
 - Solved a problem with the validation of `dataformat` plugin lang strings.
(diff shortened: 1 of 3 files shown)


stronk7:
PS: Also, added a tiny 2nd commit that updates one GH Action to newer version that doesn't emit NodeJS 16 deprecation warnings.

kabalin (approved):
Thanks Eloy, this looks good to me.

stronk7:
tata, RK!

PR #310, opened 2024-08-11
Use nvm 0.39.7 to overcome 0.40.0 regression issue

kabalin:


kabalin:
I suggest to install 0.39.7 explicitly as part of initialisation step, so we have control over the version we use. Also `ubuntu-24.04` runner image [does not have nvm](https://github.com/kabalin/moodle-plugin-ci/actions/runs/10378504275/job/28734897465), so this step will be needed sooner or later.

kabalin:
Closing in favour of #312

Issue #314, opened 2024-08-27
Presence of .scss file in a plugin makes moodle-plugin-ci grunt command to fail


marinaglancy:
We have a plugin that has a .scss file that we just store there together with a script that compiles to .css 

On moodle-plugin-ci 4.5.4 the 'grunt' command fails:

```
Run moodle-plugin-ci grunt --max-lint-warnings 0
 RUN  Grunt on block_rbreport
Running "gherkinlint" task
Done.
>> Running tasks for component directory blocks/rbreport
Running "stylelint:css" (stylelint) task
>> Linted 1 files without errors
Done.
>> Running tasks for component directory blocks/rbreport
Running "stylelint:scss" (stylelint) task
Verifying property stylelint.scss exists in config...ERROR
>> Unable to process task.
Warning: Required config property "stylelint.scss" missing. Use --force to continue.
Aborted due to warnings.
Error: Process completed with exit code 1.
```

This was not failing on moodle-plugin-ci 4.5.0

marinaglancy:
This is the plugin itself: https://github.com/moodleworkplace/moodle-block_rbreport

For now we add a workaround - the file `.moodle-plugin-ci.yml` with the contents:
```
filter:
  notPaths:
    - scss
```

kabalin:
Hm, this is likely caused by #278 (included in 4.5.1)

kabalin:
The fundamental issue is that we do not expect `scss/` directory to present in non-theme plugins when calling `grunt stylelint:scss`, so it is not added to gruntfile config.

This issue can be replicated by calling `stylelint:scss` on [block_rbreport](https://github.com/moodleworkplace/moodle-block_rbreport) repo, you will get the same error:
```
1055 rbreport:MOODLE_402_STABLE> npx grunt stylelint:scss
>> Running tasks for component directory blocks/rbreport

Running "stylelint:scss" (stylelint) task
Verifying property stylelint.scss exists in config...ERROR
>> Unable to process task.
Warning: Required config property "stylelint.scss" missing. Use --force to continue.

Aborted due to warnings.
```

[...]

PR #320, opened 2024-09-10
Use general stylelint task in grunt command

kabalin:
This change makes use of general stylelint task in grunt command, thus let grunt task to decide whether css or scss linter needs to be run (or both).

Also cleans up deprecated `stylelint:less` task (removed in 3.7 at MDL-64506) in a separate commit.

Fixes https://github.com/moodlehq/moodle-plugin-ci/issues/314


junpataleta (approved):
Thanks, @kabalin! The changes make sense and look spot-on.

PR #324, opened 2024-10-09
Recommend to use the latest version in the example usage

fulldecent:


gha.dist.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4', '8.0', '8.1']
```
gjb2048, line 33: Why have you not changed this in line with https://moodledev.io/general/releases/4.5 ?

kabalin:
Hi, this task is a part of Moodle release checklist and done only for `X.` releases.

fulldecent:
Thank you, got it. And to be clear Moodle is not following Semver here. 

So I still think this should be improved and the release checklist should be updated for all "major" versions.

PR #333, opened 2024-12-01
Update to Postgres 14 as required by main

aspark21:
plugin installs against main branch of moodle are failing since the postgres 14 requirement has been integrated

```
  Output:                                                                      
  ================                                                             
  == Environment ==                                                            
  !! database postgres (13.18 (Debian 13.18-1.pgdg120+1)) !!                   
  [System] version 14 is required and you are running 13.18 -                  
```

PR #336, opened 2025-01-28
Bump php dependencies (and other minor tweaks)

stronk7:
Bump a bunch of php dependencies to latest ones

Just update to current ones, automatically, without changing
any composer.json requirement.

Fixes #332 (and #333, just saw that one).

And, also, 3 more commits:
- Re-generates the CLI docs which features were changed by #320.  <-- (just a detail)
- Bump GHA to postgres 14, that is a requirement for Moodle 5.0.  <-- (self tests)
- Remove the PHPUnit --verbose option from Moodle >= 5.0 runs.  <-- (impacting!)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,7 +62,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:13
+        image: postgres:14
         env:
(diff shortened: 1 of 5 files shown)


kabalin (approved):
Thanks @stronk7 for housekeeping!

PR #337, opened 2025-01-31
Fix getComponentInstallDirectory plugintypes retrieval issue.

kabalin:
`fetch_plugintypes` return signature was changed at [MDL-79843](https://tracker.moodle.org/browse/MDL-79843).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](https://keepachangelog.com).
 
 ## [Unreleased]
+### Fixed
+- Fix getComponentInstallDirectory plugintypes retrieval issue to comply with 5.0
+
 ## [4.5.5] - 2024-11-06
 ### Added
 - Improvements to plugin validation implementation:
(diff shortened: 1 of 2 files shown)


PhilippImhof:
This will Fix #338; you were a split second faster than me...

paulholden (approved):
Thanks Ruslan & Philipp 👍

andrewnicols:
Hi peeps,

I'm just wondering what the ratinoale is for not using `\core\component::get_component_directory($component);`?

I know that the comment says it's beacuse you don't want to use the static cache: https://github.com/moodlehq/moodle-plugin-ci/blob/18fe3e1c9f6a94a0372904c93438523f5f011663/src/Bridge/Moodle.php#L103

But `fetch_plugintypes()` uses a static cache anyway.

The original reason for this is described in the commit message of https://github.com/moodlehq/moodle-plugin-ci/commit/6138af1b658cd05b7d5c91ca1e3e14cdc1cb9a51
> Avoid using static cache in core_component
> 
> When installing multiple plugins, core_component's
> static cache becomes stale.  This is bad when installing
> a parent plugin, then a sub-plugin.  Use reflection
> to gain access to the un-cached method we need.

If this is the case, then we should already be calling `\core\component::reset()` after placing the new plugin anyway.

Issue #351, opened 2025-05-23
Remove support for travis-ci


andrewnicols:
It's been a few years since we stopped using Travis CI in Moodle core, but we have continued to provide a travis configuration here and test it.

Travis is rarely used these days, and we currently pay to maintain it. I'd like to propose that we remove official support for it, and stop testing with it.

kabalin:
I support this idea @andrewnicols, I assume it is rarely used, we don't have stats though. I am thinking to remove test, but leave config template and docs in place, may be move it to "other CI tools" where we can add Gitlab example, but with no commitment to keep it up to date or support. We should also change documentation to imply that GHA is the only supported workflow (also given plans to add custom action #53).

kabalin:
Deprecation notes have been added to [4.5.8 release](https://github.com/moodlehq/moodle-plugin-ci/releases/tag/4.5.8).

PR #360, opened 2025-07-08
Prepare for 4.5.8 release

kabalin:
This includes Travis CI deprecation notes, effectively preparing for its removal in 5.0.0 (#351).

I am not sure if we need to output deprecation message in the tool, in theory is it possible to check Travis specific environment var, and echo message, I could not find any standard warning CI output format like GH has, would simple echo be noticed? It is also not supposed to break existing CI setup, only if users bump install to `^5` something _may_ stop working.

Any thoughts @stronk7 @andrewnicols ?

**Note for reviewer**: once approved, this needs to be merged without merge commit (use "Rebase and merge" option per [instruction](https://github.com/moodlehq/moodle-plugin-ci/blob/main/docs/ReleaseNewVersion.md).

### DIFF
--- a/.travis.dist.yml
+++ b/.travis.dist.yml
@@ -1,3 +1,4 @@
+# Travis CI support has been deprecated in 4.5.8 and will be removed in 5.0.0. Please use GitHub Actions instead.
 language: php
 
 dist: focal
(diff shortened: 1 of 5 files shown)


stronk7 (approved):
Changes (readme, changelog, binary, ... look ok for me for a normal release, unless I'm missing something, so +1, thanks!

PR #376, opened 2026-02-03
Support Moodle cloning from local repo

### LINKED ISSUE #306: Support Moodle cloning from local repo
Cloning from the local path is standard `git clone` [functionality](https://git-scm.com/docs/git-clone#_git_urls), which would be useful to support for `--repo` param. [Validation function](https://github.com/moodlehq/moodle-plugin-ci/blob/main/src/Validate.php#L93-L99) needs amendment to support `file://` and local path `^[/|\./]`.

kabalin:
Fixes #306


vmdef (approved):
Thanks @kabalin .
The changes make sense and are well documented and tested.

PR #381, opened 2026-02-16
docs: gha example should use postgres 16 and Moodle 4.5 LTS

jrchamp:
As Moodle main/5.2 requires postgres:16 and Moodle 4.5 is the current LTS, bump both values in the sample GitHub Action file.

### DIFF
--- a/gha.dist.yml
+++ b/gha.dist.yml
@@ -8,7 +8,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:15
+        image: postgres:16
         env:
           POSTGRES_USER: 'postgres'
           POSTGRES_HOST_AUTH_METHOD: 'trust'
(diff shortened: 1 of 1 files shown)


abias:
I just ran into this issue as well on MOODLE_502_STABLE and would appreciate if this PR could be merged.

92b9fa34a717cb61bca805dc841591cbeff94d7f added PG 16 to the internal tests and to the changelog, but not to gha.dist.yml as well (which is were I always look first for changes when a new Moodle major release is out).

gha.dist.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4', '8.0', '8.1']
```
kabalin:
```suggestion
        php: ['8.1', '8.2', '8.3']
```

